PR #7, opened 2014-08-27
Add crm_claim_merge, allows to merge claims.

guewen:


crm_claim_merge/crm_claim.py
```diff
@@ -0,0 +1,370 @@
...
+    'message_ids',
+    'message_follower_ids',
+]
+
+
+class crm_claim(orm.Model):
+    _inherit = 'crm.claim'
+
+    def _merge_get_default_main(self, cr, uid, claims, context=None):
+        return sorted(claims, key=attrgetter('date'))[0]
```
pedrobaeza, line 47: Why don't you call `_merge_sort` for using the same criteria?

pedrobaeza:
Some PEP8 issues:

```
./crm_claim_merge/__openerp__.py:45:1: E124 closing bracket does not match visual indentation
./crm_claim_merge/crm_claim.py:223:80: E501 line too long (80 > 79 characters)
./crm_claim_merge/crm_claim.py:254:80: E501 line too long (82 > 79 characters)
```

crm_claim_merge/crm_claim.py
guewen, line 47: You don't necessarily want the same criteria, you can override to have another "default main", but you are right, here I can call _merge_sort in the implementation by default.

guewen, line 47: I read again the code, and that's not the same thing, _merge_get_default_main is to get the claim we'll merge the others in, and _merge_sort gives the priority for the merges of the other claims, so _merge_sort receives a list where the the main claim has already been removed. I find it confusing to call _merge_sort from _merge_get_default_main because it has not the same purpose, even if by chance they have the same sorting.

guewen:
pep8 corrected

crm_claim_merge/crm_claim.py
pedrobaeza, line 47: OK, thanks for the clarification! Do you think this deserves a comment to avoid future doubts?

guewen, line 47: I improved the docstrings a bit.

pedrobaeza:
The error is unrelated, so :+1:

hbrunn:
:+1:

PR #9, opened 2015-01-22
[ADD] port lettermgmt

hbrunn:


lettermgmt/models/__init__.py
```diff
@@ -18,12 +17,10 @@
...
-###############################################################################
-
+##############################################################################
 from . import letter_class
 from . import letter_folder
 from . import res_letter
 from . import letter_type
 from . import letter_reassignment
-from . import letter_history
 from . import letter_channel
```
ehdem: Better to do:

from . import (
    letter_class,
    letter_folder,
    res_letter,
    letter_type,
    letter_reassignment,
    letter_channel,
)

lettermgmt/models/letter_folder.py
```diff
@@ -19,20 +19,19 @@
...
         'name': fields.char('Name', required=True),
-        'code': fields.char('Code', size=8, required=True),
-        'letter_ids': fields.one2many('res.letter', 'folder_id', string='Letters',
-                                      help='Letters contained in this folder.'),
+        'code': fields.char('Code', required=True),
+        'letter_ids': fields.one2many(
+            'res.letter', 'folder_id', string='Letters',
+            help='Letters contained in this folder.'),
     }
     _sql_constraints = [('code_uniq', 'unique(code)', 'Code must be unique !')]
```
ehdem, line 37: Put _('Code must be unique !') to be translated

lettermgmt/models/letter_type.py
```diff
@@ -20,17 +20,16 @@
...
     _name = 'letter.type'
-    _description = _("Letter Type")
+    _description = "Letter Type"
     _columns = {
-        'name': fields.char('Type', size=32, required=True),
-        'code': fields.char('Code', size=8, required=True),
+        'name': fields.char('Type', required=True),
+        'code': fields.char('Code', required=True),
     }
     _sql_constraints = [('code_uniq', 'unique(code)', 'Code must be unique !')]
```
ehdem, line 35: Put _('Code must be unique !') to be translated

lettermgmt/models/__init__.py
guewen: I disagree. When importing several things from the same module, it is a usual practice to group them, but import of  different modules is used to be on several lines.

lettermgmt/models/letter_folder.py
guewen, line 37: AFAIK this is not necessary on sql constraints.

hbrunn:
@guewen @ehdem okay, grouped imports it is
@ehdem constraint messages are translated automatically

hbrunn:
also here, I merged #15

__unported__/lettermgmt/letter_history.py
```diff
@@ -1,41 +0,0 @@
...
-import time
-
-
-class letter_history(orm.Model):
-    _name = "letter.history"
-    _description = _("Letter Communication History")
-    _order = "id desc"
-    _columns = {
-        'register_id': fields.many2one('res.letter', 'Register'),
-        'name': fields.char('Action', size=64),
```
StefanRijnhart, line 35: Moving the functionality to chatter seems like a good idea. Can you post any existing history items as chatter messages by means of migration?

lettermgmt/models/res_letter.py
```diff
@@ -0,0 +1,156 @@
...
+            [('in', 'IN'), ('out', 'OUT')], 'Move', readonly=True,
+            help="Incoming or Outgoing Letter."),
+        'type': fields.many2one(
+            'letter.type', 'Type',
+            help="Type of Letter, Depending upon size."),
+        'class': fields.many2one(
+            'letter.class', 'Class', help="Classification of Document."),
+        'date': fields.datetime('Letter Date', help='The letter\'s date'),
+        'snd_rec_date': fields.datetime(
+            'Sent / Received Date', required=True,
```
StefanRijnhart: This field is required now. I understand that the date received of an incoming letter can be required as it should be known in any state. But in the case of an outgoing letter, should it maybe be kept without a default and be set when setting the state to 'Sent' if it is still empty at that point?

__unported__/lettermgmt/letter_history.py
hbrunn, line 35: This code was already dead in the 7.0 version, so a migration would have to take place there

lettermgmt/models/res_letter.py
hbrunn: done

StefanRijnhart:
Thanks! :+1:

pedrobaeza:
I would put as module name only 'letter', as mgmt, management and so on seems redundant. As this can be generalized to other documents, maybe another name is also advisable, but that implies a lot of changes in all labels.

lettermgmt/__openerp__.py
```diff
@@ -22,44 +22,30 @@
 
 {
     'name': 'Letter Management',
-    'version': '0.1',
+    'version': '0.2',
```
pedrobaeza: Why not 1.0?

lettermgmt/__openerp__.py
```diff
@@ -22,44 +22,30 @@
...
     'author': 'Savoir-faire Linux',
     'maintainer': 'Savoir-faire Linux',
     'website': 'http://www.savoirfairelinux.com',
     'license': 'AGPL-3',
     'category': 'Social Network',
```
pedrobaeza: This category doesn't fit exactly, doesn't it?

hbrunn:
@pedrobaeza mind you that this a port of an existing module, so I wouldn't rename it if not absolutely necessary

lettermgmt/__openerp__.py
hbrunn: 'Customer Relationship Management' better?

lettermgmt/__openerp__.py
hbrunn: does it matter?

hbrunn: it makes sense to use 1.2 here as the code for v7 is version 1.1

lettermgmt/__openerp__.py
pedrobaeza: Yeah!

pedrobaeza:
This is the moment for the renaming, as this is the entry PR to this version, but as you want.

StefanRijnhart:
@pedrobaeza Renaming standalone community modules is expensive for its users. While we can cover the migration easily in OpenUpgrade, this is not the case for upgrades covered by the Odoo enterprise contract. Non-expert users would be left with inconsistent databases and be forced to perform an uninstall/fresh install cycle of the module, losing pre-existing data in the process. So while I agree with your claim that the current name is not optimal, I would still vote against renaming it.

pedrobaeza:
OK, as you prefer. I suggested that because the number of DBs that is already using this in v7 that we can't cover and that is going to migrate to v8 is minimal.

:+1:

Issue #10, opened 2015-01-22
lettermgmt: semantics of send_id, recipient_ids, partner_id, extern_partner_ids


hbrunn:
@bwrsandman,

I'm now busy porting lettermgmt to 8.0 and I'm a bit confused about the semantics of the fields mentioned above. In 6.0, there was snd_rec_id and partner_id which was respectively the user who sent a letter and the partner to receive it if it's an outgoing letter, while snd_rec_id was the addressee and partner_id the sender if it's an incoming letter.

Now in 7.0, the incoming letter form gives me partner_id as recipient and send_id as sender. I don't understand the reasoning here, shouldn't it be the other way around, that I send something to a partner and the sender necessarily is a user?

Then for outgoing letters, we have recipient_ids (which refers to res.users) as recipients and no sender. There's also extern_partner_ids what I would have expected as the recipients field.

What do I misunderstand here?

bwrsandman:
It's probably a mistake on my part. We ended up not using `lettermgmt`.

It's been a while since I touched `lettermgmt`, it might not have fully been ported.

hbrunn:
thanks @bwrsandman, so if I PR some changes to 7.0 following the lines above you wouldn't mind?

bwrsandman:
Go for it!

pedrobaeza:
Can this be closed?

hbrunn:
yes

PR #12, opened 2015-01-23
[DEL] unmaintained lettermgmt_hr

hbrunn:
This module does not even install and is unmaintained as per #10


hbrunn:
travis fails because of pylint errors in newsletter

yvaucher:
@bwrsandman @max3903 agreed on this? Or any plan to make it installable?

yvaucher:
According to https://github.com/OCA/crm/issues/10#issuecomment-71084944

:+1:

pedrobaeza:
I would put it on `__unported__` instead, an remove it in v8.

hbrunn:
@pedrobaeza this module is new, so I don't really understand your proposal?

pedrobaeza:
OK, I think I get it, this version is the 6.1 one, and you are going to rename it to lettermgmt. Doesn't it?

bwrsandman:
:+1:

yvaucher:
@pedrobaeza No this one extends `lettermgmt`

pedrobaeza:
And the source is from 6.1? If so, this should be moved to `__unported__`, not removed.

bwrsandman:
@pedrobaeza If I recall correctly, lettermgmt from 6.1 wasn't an OCA module and a lot of cleaning was needed to bring it to our standards. At this point, the old 6.1 lettermgmt would need to be in `__unported__` for 6.1 and that makes no sense.

pedrobaeza:
OK, thanks for the explanation.

PR #13, opened 2015-02-02
[IMP] autofix conventions

hbrunn:


tafaRU:
LGTM :+1: , thanks!

crm_lead_stage_monitoring/crm.py
```diff
@@ -35,9 +35,10 @@ def write(self, cr, uid, ids, vals, context=None):
             vals['last_activity_stage'] = fields.datetime.now()
         return super(CrmLead, self).write(cr, uid, ids, vals, context=context)
 
-    def copy(self, cr, uid, id, default=None, context=None):
+    def copy(self, cr, uid, rec_id, default=None, context=None):
```
rvalyi, line 38: I don't remember the thread exactly but I think there has been some questioning if we could change the name of id in the signature of default ORM methods because of some introspection happening in the core expecting just "id". I think this was discussed when somebody wanted to test for variable named "id" in the automatic tests in the OCA. Am I wrong with this or has the debate been closed since? Anybody remember?  @pedrobaeza @StefanRijnhart @bwrsandman any idea?

PR looks LGTM other than this.

yvaucher, line 38: Well spoted.

It seams that could have some side effect with old/new API compatibility when guessing how it needs to be decorated:
https://github.com/odoo/odoo/blob/8.0/openerp/api.py#L641
So I think we can set `rec_id` but we need to specify its decorator:
`@api.cr_uid_id_context`

Or we can revert to `id`

pedrobaeza, line 38: Yeah, that's right. We deactivate also this check on MQT because this problem.

StefanRijnhart, line 38: Ugh! The new API encourages the infamous use of a reserved word as a variable name :broken_heart:

Please use the decorator instead of disabling this style check. BTW what is MQT exactly?

yvaucher, line 38: @StefanRijnhart My guess: MQT > Maintainer Quality Tools

StefanRijnhart, line 38: Thanks @yvaucher. @Pedrobaeza, I see MQT contains the master config for the OCA checks. Where is this check disabled exactly?

yvaucher, line 38: @StefanRijnhart https://github.com/OCA/maintainer-quality-tools/pull/136

StefanRijnhart, line 38: Thanks @yvaucher. I see overwhelming support for the change on that thread so I'll just leave it then. But I do prefer the explicit decorator solution.

StefanRijnhart:
Setting to needs fixing for the missing decorator, otherwise LGTM

crm_lead_stage_monitoring/crm.py
StefanRijnhart, line 38: NB. this is for 7.0

StefanRijnhart:
Giving my support anyway because the introspection bug does not affect 7.0. :+1:

yvaucher:
Indeed :+1:

StefanRijnhart:
For those concerned about the use of the 'id' keyword in decorator guessing (again, not relevant for this 7.0 code): my tentative solution for this is already merged in 8.0 (although I opted for res_id akin to act window definitions, not rec_id): https://github.com/odoo/odoo/pull/5027/files.

pedrobaeza:
Thanks for the information. Good to see that there's an alternative, although it's still a hardcoded one.

yvaucher:
@StefanRijnhart Thanks, good to know we can use `res_id` now.

PR #15, opened 2015-02-06
[UPD] .travis.yml from maintainer-quality-tools

hbrunn:
I already added the dependency to web because newsletter needs it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ env:
   - VERSION="8.0" LINT_CHECK="1"
   - VERSION="8.0" ODOO_REPO="odoo/odoo" LINT_CHECK="0"
   - VERSION="8.0" ODOO_REPO="OCA/OCB" LINT_CHECK="0"
-  - VERSION="8.0" ODOO_REPO="odoo/odoo" UNIT_TEST="1" LINT_CHECK="0"
 
 virtualenv:
   system_site_packages: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
   - "2.7"
 
 env:
-  - LINT_CHECK="1"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,46 +1,26 @@
...
+  - "2.7"
 
 env:
-    - ODOO="https://github.com/savoirfairelinux/odoo/archive/setuptools-addons.tar.gz"  # Temp until https://github.com/odoo/odoo/issues/185 or https://github.com/odoo/odoo/issues/441 is fixed
-#    - ODOO="https://github.com/odoo/odoo/archive/master.tar.gz"
-#    - ODOO="https://github.com/OCA/OCB/archive/master.zip"
+  - LINT_CHECK="1"
+  - VERSION="8.0" ODOO_REPO="odoo/odoo" LINT_CHECK="0"
+  - VERSION="8.0" ODOO_REPO="OCA/OCB" LINT_CHECK="0"
+  - VERSION="8.0" ODOO_REPO="odoo/odoo" UNIT_TEST="1" LINT_CHECK="0"
```
pedrobaeza: This is very slow when the number of modules are medium-high

hbrunn: To be honest, I'm not sure why we need that when we have the two lines before. That's how it's proposed in https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml, but I'm happy to remove this if you agree

pedrobaeza: It was proposed to make isolated test for each module to check dependencies between them. You can remove it with no problem.

yvaucher:
:+1:

pedrobaeza:
:+1:

dreispt:
:+1:

rvalyi:
:+1:

PR #17, opened 2015-02-18
8.0 [ADD] Module crm_lead_sale_link

ghost:
This module adds buttons in the CRM Opportunity form to access the quotations related to the partner.


crm_quotation_link/__openerp__.py
```diff
@@ -0,0 +1,55 @@
...
+##############################################################################
+
+{
+    'name': 'CRM Quotation Link',
+    'version': '1.0',
+    'author': 'Savoir-faire Linux',
+    'license': 'AGPL-3',
+    'category': 'Others',
+    'summary': 'CRM Quotation Link',
+    'description': """
```
pedrobaeza: This is not needed due to the README file.

crm_quotation_link/README.rst
```diff
@@ -0,0 +1,52 @@
...
+To use this module, you need to :
+- create a lead
+- transform it into an opportunity
+- create a quote from this opportunity by using the button
+- you should get back and keep link beetween the models
+
+Known issues / Roadmap
+======================
+
+Contributors
```
pedrobaeza: Contributors section goes under Credits and with `----` style. The same for Maintainer.

crm_quotation_link/models/crm_lead.py
```diff
@@ -0,0 +1,75 @@
...
+from openerp import models, fields, api, _
+
+
+sales_order_states = [
+    'progress', 'manual', 'shipping_exept', 'invoice_except', 'done']
+
+quotations_states = ['draft', 'sent', 'waiting_date']
+
+
+class crm_lead(models.Model):
```
pedrobaeza: Use CamelCase for class name

crm_quotation_link/models/crm_lead.py
```diff
@@ -0,0 +1,75 @@
...
+    _inherit = 'crm.lead'
+
+    def count_sales_order(self):
```
pedrobaeza: You have to set a one2many to sale.order, and make @api.depends on that field. Calculation will be `len(field.filtered())`

crm_quotation_link/models/crm_lead.py
```diff
@@ -0,0 +1,75 @@
...
+    quotations_count = fields.Integer(compute='count_sales_order')
+
+    @api.multi
+    def get_sale_order_view(self, order_state, view_title):
+        partner_ids = list({
+            lead.partner_id.id for lead in self
+            if lead.partner_id
+        })
+
+        return {
```
pedrobaeza: You can expand this to show directly the form if there is only one record.

pedrobaeza:
A better name would be `crm_lead_sale_link`

ghost:
@bjoliveau What about crm_lead_sale_link instead of crm_quotation_link?

crm_quotation_link/models/crm_lead.py
ghost: @pedrobaeza I don't think this is required. At least, it is not required for the purpose of this module. The field sales_order_count is only used to display the quantity of sales orders on the button in the view.

pedrobaeza: But if you want correct data, you have to do that, because cache mechanism will not be alerted when new orders are added.

ghost: I tried it, went in the opportunity form. The value was 2. Then, I created a new sale order. Went a second time in the opportunity form. The value is now 3. In what case would it mess with the cache?

pedrobaeza: Have you tried from another user in other browser?

ghost: Just tried it, the same thing happens.

ghost:
@pedrobaeza Added your logo here as well.

pedrobaeza:
:+1:

crm_lead_sale_link/README.rst
```diff
@@ -0,0 +1,56 @@
+CRM Lead Sale Link
+==================
+
+This module extends the functionality of crm to deliver two contexted buttons into the opportunity form view :
+    the first opens all quotes linked to this opportunity
+    the second opens all sales order linked to this opportunity
```
rdeheele: s/sales order/sales orders

crm_lead_sale_link/README.rst
```diff
@@ -0,0 +1,56 @@
...
+No configuration to use this module.
+
+Usage
+=====
+
+To use this module, you need to :
+- create a lead
+- transform it into an opportunity
+- create a quote from this opportunity by using the button
+- you should get back and keep link beetween the models
```
rdeheele: s/beetween/between

crm_lead_sale_link/README.rst
```diff
@@ -0,0 +1,56 @@
...
+Contributors
+------------
+
+David DUFRESNE <david.dufresne@savoirfairelinux.com>
+Jordi RIERA <jordi.riera@savoirfairelinux.com>
+Bruno JOLIVEAU <bruno.joliveau@savoirfairelinux.com>
+
++Icon
++----
++* http://en.wikipedia.org/wiki/File:People%27s_Action_Party_of_Singapore_logo.svg
```
rdeheele: The icon is not displayed, and I'm not sure about free usage.

rdeheele:
few little remarks, :+1:

crm_lead_sale_link/README.rst
pedrobaeza: This reference is incorrect (it belongs to another module, although the concerns about its free usage are true). Please remove it.

hbrunn:
:+1:

PR #19, opened 2015-02-19
Add Module crm_actions

ghost:


pedrobaeza:
Please use singular for the module name for keeping homogeneity.

crm_actions/README.rst
```diff
@@ -0,0 +1,54 @@
...
+===========
+
+This module was written to extend crm features.
+It delivers new object named "Actions" to follow history around leads and opportunites.
+This new object depends on ACL groups you give to users.
+
+Installation
+============
+
+To install this module, you just need to select the module and insure yourself dependancies are available.
```
pedrobaeza: s/dependancies/dependencies

pedrobaeza:
For not getting confused with the module icon, as it is the same as the crm one, you can overlay another image/icon to refer that this is a submodule with a more specific option. This icon for example fits well: http://en.wikipedia.org/wiki/File:People%27s_Action_Party_of_Singapore_logo.svg

You can take these icons as example:
- Main: https://github.com/odoomrp/odoomrp-wip/blob/8.0/quality_control/static/description/icon.png
- Submodule: https://github.com/odoomrp/odoomrp-wip/blob/8.0/quality_control_mrp/static/description/icon.png

ghost:
@pedrobaeza Thanks for your review. I removed the icon, if later I find one that could be used, I will add it.

pedrobaeza:
@dufresnedavid, I have made you a PR with an icon.

pedrobaeza:
:+1:

crm_action/models/crm_action.py
```diff
@@ -0,0 +1,77 @@
...
+        lead = self.lead_id
+        if lead and lead.partner_id:
+            self.partner_id = lead.partner_id.id
+
+    partner_id = fields.Many2one(
+        'res.partner', string='Customer')
+
+    date = fields.Date(
+        'Date', required=True,
+        default=lambda self: fields.Date.context_today(self))
```
hbrunn: you can achieve the same by saying `default=fields.Date.context_today`

crm_action/models/crm_action.py
```diff
@@ -0,0 +1,77 @@
...
+
+    user_id = fields.Many2one(
+        'res.users', string='User', required=True,
+        default=lambda self: self.env.user)
+
+    def search_action_types(self):
+        return self.env['crm.action.type'].search(
+            [('is_active', '=', True)]).sorted(lambda a: a.priority)
```
hbrunn: `operator.itemgetter('priority')` is considered more pythonic by some people.

But why don't you pass 'priority' as the order argument to search in this case?

hbrunn:
a few nags, for the rest: :+1:

crm_action/README.rst
```diff
@@ -0,0 +1,58 @@
...
+- you can overview all actions for any lead or opportunity with the new menu entry
+
+Known issues / Roadmap
+======================
+
+Credits
+=======
+
+Module developed and tested with Odoo version 8.0
+For questions, please contact our support services <support@savoirfairelinux.com>
```
StefanRijnhart: Is this kind of sales redirection acceptable in an OCA module? This raises all kinds of questions with regards to shared responability and co-ownership between the current owners and any future maintainers. Would you consider removing this line?

StefanRijnhart:
I do appreciate everyone's effort to make nice icons, but in this case I don't think that applying the logo of a political party is a good idea.

crm_action/README.rst
pedrobaeza: I also agree that we should keep neutrality in OCA modules, so I vote for removing this line.

StefanRijnhart:
Tested. Should work well, on the whole.

One usability issue: users that have permission on own leads only can see actions related to other leads. Can you align these permissions by adding record rules for the standard CRM groups on this action model?

And a feature request: how about setting the filter on 'My actions' active by default? Again, this aligns with the views on leads and prospects, and prevents this view to get cluttered too quickly when working with multiple salespersons.

pedrobaeza:
Sorry, it was my fault about the icon. I didn't see the disclaimer, but the aesthetics only.

@dufresnedavid, please remove it and I'll find later a better icon.

crm_action/__openerp__.py
```diff
@@ -0,0 +1,45 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'CRM Action',
+    'version': '1.1',
+    'author': 'Savoir-faire Linux',
```
StefanRijnhart: Please add OCA as an author, for apps store visibility of the OCA. See https://github.com/OCA/crm/blob/8.0/newsletter/__openerp__.py#L25 for an example.

StefanRijnhart:
@pedrobaeza That's alright. Glad you agree!

ghost:
Thanks @StefanRijnhart Implemented your suggestions!

StefanRijnhart:
Thanks. I miss one thing: the record rules filtering out records based on lead_id.user_id:
['|',('lead_id.user_id','=',user.id),('lead_id.user_id','=',False)] for group base.group_sale_salesman and [(1,'=',1)] for group base.group_sale_salesman_all_leads. See 
addons/crm/security/crm_security.xml from the core distribution.

ghost:
@StefanRijnhart Thanks, I added the rules.

StefanRijnhart:
Thanks. You chose for the user that is directly linked to the lead, but that value can be changed to a different user in your data model. That could lead to some inconsistency: it would be possible to assign the user to a lead that he cannot access, and he would not be able to check actions from his own links but assigned to someone else. How would that have to be managed properly?

ghost:
@StefanRijnhart One solution is to allow the user to see both all actions related to his own leads and all actions assigned to him.

StefanRijnhart:
Yes, that would be a pragmatic solution.

ghost:
@StefanRijnhart I fixed the access rule as discussed.

StefanRijnhart:
Thanks! :+1:

foutoucour:
:+1:

hbrunn:
@dufresnedavid the last comment to address is https://github.com/OCA/crm/pull/19#discussion_r26904389, after that, I'll be glad to proceed with the merge

ghost:
@hbrunn Thanks, it's done !

bwrsandman:
:+1:

PR #21, opened 2015-03-05
[ADD] manage lost reason improvements

rdeheele:


crm_lead_lost_reason/wizard/lost_reason.py
```diff
@@ -44,3 +44,14 @@ def confirm_lost(self):
         lead.lost_reason_id = self.reason_id.id
         lead.case_mark_lost()
         return act_close
+
+    @api.model
+    def default_get(self, field_names):
```
pedrobaeza: Instead of overwriting default_get, you can also put a specific method in default attribute of the field, to isolate a bit more.

rdeheele: you're right, it's better, thanks Pedro

pedrobaeza:
:+1:

crm_lead_lost_reason/wizard/lost_reason.py
```diff
@@ -28,10 +28,17 @@ class CrmLeadLost(models.TransientModel):
     _name = 'crm.lead.lost'
     _description = __doc__
 
+    def _default_reason(self):
+        active_id = self._context.get('active_id', [])
+        if active_id:
+            lead = self.env['crm.lead'].browse(active_id)
```
gurneyalex, line 35: I'd check that context['active_model'] == 'crm.lead' just to be on the safe side.

gurneyalex, line 35: not sure about the rules regarding active_id vs active_ids in the context

crm_lead_lost_reason/wizard/lost_reason.py
```diff
@@ -28,10 +28,17 @@ class CrmLeadLost(models.TransientModel):
     _name = 'crm.lead.lost'
     _description = __doc__
 
+    def _default_reason(self):
+        active_id = self._context.get('active_id', [])
```
gurneyalex: you can remove the second argument of your call to get(): it defaults to `None`, which is `False`-ish

lepistone: `, []` is not needed.

crm_lead_lost_reason/wizard/lost_reason.py
rdeheele, line 35: @gurneyalex thanks, ok it's safer to check the active_model.
in the context, the user marks lost leads one by one.

crm_lead_lost_reason/wizard/lost_reason.py
rdeheele: @gurneyalex @lepistone thanks, done

rdeheele:
I allowed myself to add another change under this PR.

pedrobaeza:
:+1: for this too.

gurneyalex:
:+1:

lepistone:
:+1: thanks!

PR #22, opened 2015-03-17
[Add] crm_reason_to_export

ghost:


nhomar:
The field cannot be called x_\* it will be mixed with the customized ones thorught the interface.

I do not understand functionally speaking why do I need to explain or set a "Reason to Export" in a sale order, if you can extend such explanation on the README, it can help everybody to re-use this module and mix with them applications.,

Regards.

PR #24, opened 2015-04-08
[ADD] Membership withdrawal addon

antespi:
# Membership withdrawal

This module was written to extend the functionality of membership to support
tracking of members signup and withdrawal. Also, allows to select a withdrawal
reason from a configurated list of reasons


membership_withdrawl/README.rst
```diff
@@ -0,0 +1,63 @@
...
+fields are not related with Membership Lines or Membership State
+
+
+Known issues / Roadmap
+======================
+
+None
+
+
+Credits
```
dreispt: Should the icon author be credited?

antespi: Yes, you're right, sorry.
* Original source : https://thenounproject.com/term/add-group/79375 by Mundo (https://thenounproject.com/DMundo), 2014
* Modified by Antonio Espinosa using GIMP image editor, 2015

antespi:
Any new reviews? Thanks

membership_withdrawl/__openerp__.py
```diff
@@ -0,0 +1,40 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': "Membership withdrawal",
+    'category': 'Association',
+    'version': '1.0',
+    'depends': [
+        'membership',
```
pedrobaeza: In fact, this dependency is not needed, because you are not using nothing of it. You can even generalise and call it _partner_withdrawal_.

antespi: There is a dependency in res_partner_view.xml view:

```
<field name="inherit_id" ref="membership.view_partner_form"/>
```

Maybe this XML file is not named correctly? This view inherit from a view that inherits from res_partner_view, declared in membership. If module 'membership is not installed, this view will not find `associate_member` field:

```
<field name="associate_member" position="after">
```

pedrobaeza: Yeah, I see. Well, you can practically reference the parent view instead the membership one, losing only the "section" to put the field into. As you like, but the other approach can be applied to other things, not just membership, but partnerl withdrawal in general.

antespi:
@pedrobaeza how can i change the name of this addon?, closing this PR and open a new one or just change folder name?

pedrobaeza:
`git mv` in your local branch, make a commit, and push again into this remote branch.

partner_withdrawal/i18n/es.po
```diff
@@ -0,0 +1,83 @@
...
+"POT-Creation-Date: 2015-04-08 12:20+0000\n"
+"PO-Revision-Date: 2015-04-08 12:20+0000\n"
+"Last-Translator: <>\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: membership_withdrawl
```
hbrunn: you should regenerate those files after the module rename, otherwise you won't have translations

partner_withdrawal/i18n/membership_withdrawl.pot
```diff
@@ -0,0 +1,83 @@
...
+"POT-Creation-Date: 2015-04-08 12:20+0000\n"
+"PO-Revision-Date: 2015-04-08 12:20+0000\n"
+"Last-Translator: <>\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: membership_withdrawl
```
hbrunn: see above

hbrunn:
after the translations are fixed, :+1:

antespi:
@hbrunn you're right, thanks a lot. Change module name implies lots of changes ;), now i realize.
I've just change this

jgrandguillaume:
:+1:

PR #28, opened 2015-06-03
[ADD] new addon project_recalculate

Endika:
# Project recalculate

This module allows you to recalculate a tasks project, assign "date from" and "estimation days" in project tasks, and runs recalculate proyect button in project profile. The tasks recalculate according a new field "Calculation type".

If the project date is changed or modified in the gant view, then all the pending tasks are retro-planned again according the new date.
# Usage

- First step, create a project or project template and assign calculation type and date begin or date end depending for calculation type.
- Second step, indicate in to tasks of project the "from days" and "estimation days".
- And Last step, runs "recalculate project" in project profile.


dreispt:
This PR should be for the `project-service` repo.

Endika:
new PR https://github.com/OCA/project-service/pull/90

PR #32, opened 2015-07-01
[FIX] don't expect a singleton when searching for

hbrunn:
recipient objects

### DIFF
--- a/newsletter/model/newsletter_newsletter.py
+++ b/newsletter/model/newsletter_newsletter.py
@@ -68,7 +68,7 @@ class newsletter_newsletter(models.Model):
                 'template_id': self.type_id.email_template_id.id,
                 'default_res_id': self.id,
                 'newsletter_res_id': self.env[self.type_id.model.model].search(
-                    safe_eval(self.type_id.domain)).ids[-1] or False,
+                    safe_eval(self.type_id.domain)).ids[:-1] or False,
             },
             'view_mode': 'form',
             'view_id': self.env.ref(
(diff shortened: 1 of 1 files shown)


newsletter/model/newsletter_newsletter.py
```diff
@@ -68,7 +68,7 @@ def action_preview(self):
                 'template_id': self.type_id.email_template_id.id,
                 'default_res_id': self.id,
                 'newsletter_res_id': self.env[self.type_id.model.model].search(
-                    safe_eval(self.type_id.domain)).id or False,
+                    safe_eval(self.type_id.domain)).ids[-1] or False,
```
pedrobaeza: It's better to make `safe_eval(self.type_id.domain))[:1].id` so you handle all cases (even empty recordset)

hbrunn:
@pedrobaeza thanks, much better to use a slice here

guewen:
:+1:

oihane:
:+1:

newsletter/model/newsletter_newsletter.py
```diff
@@ -68,7 +68,7 @@ def action_preview(self):
                 'template_id': self.type_id.email_template_id.id,
                 'default_res_id': self.id,
                 'newsletter_res_id': self.env[self.type_id.model.model].search(
-                    safe_eval(self.type_id.domain)).id or False,
+                    safe_eval(self.type_id.domain)).ids[:-1] or False,
```
pedrobaeza, line 71: This is returning the last element, not the first, which can be semantically different (not sure about this case). That's why I told you to put `.ids[:1]`

hbrunn, line 71: @pedrobaeza it's arbitrary. we just need to pick some record for the preview, and i figured the last object is nicer in some cases. But this totally depends on the sort order of said object and doesn't matter in the end

pedrobaeza, line 71: :+1: then

PR #36, opened 2015-07-08
[ADD] mail_restrict_follower_selection

hbrunn:
# Restrict follower selection

This module was written to allow you to restrict the selection of possible followers. For example, if you use the social ERP functions only internally, it makes sense to filter possible followers for being employees. Otherwise, you'll get a quite crowded list of partners to choose from.
# Configuration

To configure this module, you need to go to `System parameters` and adjust `mail_restrict_follower_selection.domain` as you see fit. This restricts followers globally, if you want to restrict only the followers for a certain record type (or have different restrictions for different record types), create a parameter `mail_restrict_follower_selection.domain.$your_model`.

Note: This module won't change existing followers!


pedrobaeza:
This should be in server-tools repo, not here, as you can use followers and so on without crm.

hbrunn:
@pedrobaeza every time i can think of some other repo than server-tools, i'd use that, because server-tools already is very cowded. crm for followers makes sense to me, let's see what other people think

pedrobaeza:
Yeah, I also think server-tools is too heavy, but crm is not also the best. Maybe it's the moment to propose a mail/social repo for this and other modules like https://github.com/OCA/server-tools/pull/198, https://github.com/OCA/server-tools/tree/8.0/mail_environment and so on?

alejandrosantana:
What about a new project?:
'messaging' encompassing anything related to email, im_chat, and internal communication, like followers.

alejandrosantana:
Oh, just what @pedrobaeza said before, I submitted without reading the last answer.

dreispt:
+1 for a new repo.
`social` or `messaging` sound good to me.

mail_restrict_follower_selection/models/mail_wizard_invite.py
```diff
@@ -0,0 +1,49 @@
...
+    @api.model
+    def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
+                        submenu=False):
+        result = super(MailWizardInvite, self).fields_view_get(
+            view_id=view_id, view_type=view_type, toolbar=toolbar,
+            submenu=submenu)
+        arch = etree.fromstring(result['arch'])
+        for field in arch.xpath('//field[@name="partner_ids"]'):
+            field.attrib['domain'] = self\
+                .mail_restrict_follower_selection_get_domain()
```
dreispt, line 47: I like this technique to split a long line. Didn't know you could do it like that!

dreispt:
This solves a pain I have, thank you.

There is another place where this feature should be added: on the mail.compose.message wizard, that you can open when adding a new message, and where you can pick Partners that will be added as Followers. 

In my case, I would like to limit possible contacts to only those that have an email address.

hbrunn:
@alejandrosantana  @dreispt @pedrobaeza https://github.com/OCA/social/pull/1

PR #38, opened 2015-07-31
[ADD] crm_claim_ref_smartbutton: Smart-button for referenced claims

pedrobaeza:
# Smart-button for referenced claims

This module adds an smart-button in the models that can be referenced in a
claim with the number of linked claims and a shortcut to them.
# Usage

Go to any record that can be referenced in a claim, and you will see in a
smart-button called _linked claims_ a number indicating the number of
claims that reference this record. Clicking on the button, it will navigate
to a list of these claims.

If you click _New_ on the claim list, the created claim will be automatically
linked to the referenced record.
# Known issues / Roadmap
- The check for seeing if the opened model is one of the possible referenced
  models is done on each view request, so this hits very little, but a bit,
  the overall performance of the system.


pedrobaeza:
@StefanRijnhart, I think you would enjoy this piece of code.

anajuaristi:
Tested functionally on runbot, it works
:+1:  RTM (ready to merge) from my functional point of view. 
Very usefull module

oihane:
:+1:
Although coveralls is failing.

pedrobaeza:
There's an error opening "My dashboard" menu entry that it's solved in this PR to Odoo: https://github.com/odoo/odoo/pull/7855

pedrobaeza:
Patch merged on Odoo, so everything is OK!

pedrobaeza:
Added tests!

oihane:
@pedrobaeza tests are failing

crm_claim_ref_smartbutton/tests/test_crm_claim_ref_smartbutton.py
```diff
@@ -0,0 +1,52 @@
...
+    def test_fields_view_get(self):
+        # Test the smart-button existence for a possible linked model
+        model_name = self.env['res.request.link'].search([])[0].object
+        res = self.env[model_name].fields_view_get()
+        eview = etree.fromstring(res['arch'])
+        xml_field = eview.xpath("//field[@name='crm_claim_linked_count_tech']")
+        self.assertTrue(
+            xml_field, "Smart-button not added to %s view" % model_name)
+        # Test the smart-button existence for a model not linked
+        for model in self.env['ir.model'].search():
```
oihane: Don't you have to put `[]` in order to search every model?

pedrobaeza:
Sorry, unfinished code was submitted.

crm_claim_ref_smartbutton/__openerp__.py
```diff
@@ -0,0 +1,25 @@
...
+              "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
+              "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/crm",
+    "contributors": [
+        "Oihane Crucelaegui <oihanecrucelaegi@avanzosc.es>",
+        "Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>",
+    ],
+    "category": "Customer Relationship Management",
+    "installable": True,
+    "pre_init_hook": "pre_init_hook",
```
lmignon: @pedrobaeza  hooks are only called at installation time not each time the server start. Therefore your monkey-patch will be only applied once. IOW your patch will no more be applied after the first time your sever will be restarted after the installation. https://github.com/odoo/odoo/commit/4105b5f0285821d8056c721b5fd50b9cae902988#diff-9f65b6631806b04e785c2acdf3854285R151

lmignon: @pedrobaeza The solution is may be to create a fake model and implement the _register_hook method to apply your monkey patch each time the module is loaded in the registry at startup time. https://github.com/odoo/odoo/blob/8.0/openerp/modules/loading.py#L446

pedrobaeza: You're only seeing half of the solution. This is the possible scenarios:
- You can start without the module copied in the addons path before the server starts.
- Or have it copied in addons path and start the server. This code prevents the monkey-patching: https://github.com/OCA/crm/pull/38/files#diff-df883f43bfa35a2a786daa20cc741e74R103
- When you install the module, in one or another case, the monkey-patching is done via the pre-hook script (https://github.com/OCA/crm/pull/38/files#diff-ec1e94af671fc87f2d489b43471edae2R14).
- Next time you start the server, the same line that previously prevents the monkey-patching, now does it (https://github.com/OCA/crm/pull/38/files#diff-df883f43bfa35a2a786daa20cc741e74R103).
- When you uninstall the module, the monkey-patching is undone here: https://github.com/OCA/crm/pull/38/files#diff-ec1e94af671fc87f2d489b43471edae2R20

lmignon: @pedrobaeza Sorry for the noise.. haven't see the full solution. IMO, the solution can be simplified by monkey-patching in a  _register_hook BaseModel in a fake non stored Model. In this way we no more need the pre_init_hook and monkey patching is done in one place.

crm_claim_ref_smartbutton/models/base_model.py
```diff
@@ -0,0 +1,105 @@
...
+def count_linked_claim_tech(self):
+    self.crm_claim_linked_count_tech = self.env['crm.claim'].search_count(
+        [('ref', '=', '%s,%i' % (self._model, self.id))])
+
+
+# Monkey patch functions - Only if module is installed
+BaseModel.action_open_linked_claims_tech = _open_related_claims
+BaseModel.count_linked_claim_tech = count_linked_claim_tech
+if 'crm_claim_ref_smartbutton' in openerp.modules.module.loaded:
+    BaseModel.base_fields_view_get = base_fields_view_get
```
lmignon: Are we sure that a python process in Odoo is dedicated to a specific database? If it's not the case and your server manage a lots of databases, chances are great that the monkey patch will be applied to all databases.

crm_claim_ref_smartbutton/__openerp__.py
pedrobaeza: Can you draft the proposition with the fake model? Take into account that this module can cover any linkable model in claims.

crm_claim_ref_smartbutton/models/base_model.py
pedrobaeza: I haven't checked in this multi-DB scenario, but there are chances than the load will be global for all DBs. It's the same problem as many of the processes of Odoo itself.

StefanRijnhart: Good point @lmignon. @pedrobaeza it is safer to check for this module being installed in the monkeypatched method, and if it is not installed simply call the original method.

As for only applying the patch if the module is loaded, you could consider using a dummy model's _register_hook (see https://github.com/Therp/Therp-Addons/blob/7.0/mail_error/model/mail_mail_monkeypatch.py#L96 for an example).

crm_claim_ref_smartbutton/__openerp__.py
lmignon:
``` python
class CrmClaimMonkeyPatcher(models.TransientModel):
    _name = 'cr.claim.mockey.patcher'

    _auto = False  # doesn't create table in database

    def _register_hook(self, cr):
        BaseModel.base_fields_view_get = base_fields_view_get
        BaseModel.fields_view_get = fields_view_get_crm_claim
```

crm_claim_ref_smartbutton/models/base_model.py
lmignon: Therefore we should check if the module is in the current context in the function `fields_view_get_crm_claim`

``` python
@api.model
def fields_view_get_crm_claim(
        self, view_id=None, view_type='form', toolbar=False, submenu=False):
    """This method should have old API signature because keyword arguments
    doesn't respect expected order (context argument at the end).
    """
    res = base_fields_view_get(
        self, view_id=view_id, view_type=view_type, toolbar=toolbar,
        submenu=submenu)
    if 'crm_claim_ref_smartbutton' not in openerp.modules.module.loaded:
        return res
    ...
```

pedrobaeza:
I have followed @lmignon's advice and implement the monkey-patching in a model with _auto=False and now the monkey-patching is only done when the module is installed, and now is multi-db aware thanks to a list that registers the dbs where the module is installed.

pedrobaeza:
@antespi you can check now

crm_claim_ref_smartbutton/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Smart-button for referenced claims
+==================================
+
+This module adds an smart-button in the models that can be referenced in a
```
gurneyalex: s/ an / a /

crm_claim_ref_smartbutton/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Smart-button for referenced claims
```
gurneyalex, line 5: format as document title

crm_claim_ref_smartbutton/models/base_model.py
```diff
@@ -0,0 +1,113 @@
...
+
+# This should be monkey-patched always, or the computed field won't be
+# added to all models due to the module loading order
+models.BaseModel._add_magic_fields = _add_magic_fields_wrapper(
+    models.BaseModel._add_magic_fields)
+models.BaseModel.count_linked_claim_tech = count_linked_claim_tech
+
+
+class CrmClaimMonkeyPatcher(models.Model):
+    _name = 'cr.claim.monkey.patcher'
```
gurneyalex: Nit picking, but still, as this got me puzzled: please s/cr/crm/ ( and update in ir.model.access.csv)

pedrobaeza:
@gurneyalex, changes done. Thanks for the review.

crm_claim_ref_smartbutton/__init__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+# (c) 2015 Pedro M. Baeza
+# License AGPL-3 - See LICENSE file on root folder for details
+##############################################################################
+
+from openerp.models import BaseModel
+from openerp.addons.crm_claim_ref_smartbutton.models.base_model import (
```
antespi: Couldn't you use

```
from .models.base_model import registered_dbs
```

here?

antespi:
:+1: tested in runbot
This is a cool monkey-patching sample. I take note because i'll need it in the future for some coming addons

crm_claim_ref_smartbutton/__init__.py
pedrobaeza: Yeah, it's better. Thanks for the tip.

pedrobaeza:
All remarks are solved, so this is ready to be merged.

hbrunn:
@pedrobaeza nice solution. But the full table scan on `res_request_link` for every view load (can't this be ormcached?) bothers me, as well as the unconditional monkey patches.

How about monkey patching only existing models' `fields_view_get` that show up in `res.request.link`, and also monkey patch https://github.com/OCA/OCB/blob/8.0/openerp/modules/registry.py#L486 in `_register_hook` in order to collect new models popping up which are in `res.request.link`? Then you can do the work in `fields_view_get` unconditionally.

Then some details like adding/removing the patch in `res.request.link#create/unlink`

rafaelbn:
@pedrobaeza I think you should answer @hbrunn and I will make a test in runbot in order to merge. OK?

rafaelbn:
Tested un runbot :+1:

anajuaristi:
Hi, could we merge this one? I think it could be very usefull have it available to use instead of staying as PR.

pedrobaeza:
@anajuaristi, there's already some suggestions on how to implement certain things that I haven't applied yet due to lack of time, so better to stay as a PR yet.

StefanRijnhart:
Agree with @hbrunn mainly about the monkeypatching (that will 'affect' instances with this module sitting uninstalled in the module path. Setting to WIP as per @pedrobaeza's last comment.

StefanRijnhart:
BTW. can a project admin create milestones for the older releases, so that we don't have to use the tag?

pedrobaeza:
cc @Tecnativa

crm_claim_ref_smartbutton/models/base_model.py
```diff
@@ -0,0 +1,111 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Pedro M. Baeza
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import fields, api, models, _
+from lxml import etree
+
+
+base_fields_view_get = models.BaseModel.fields_view_get
```
nhomar, line 9: Why the monkeypatch?

Did you know you can inherit BaseModel Normally right?

BTW, due to this has an static folder this code will try to be executed.

Can you try do the feature without monkeypatch?

nhomar:
IOH: This kind of changes without tests (changes that modifiy the general behavior of odoo) are really dangerous, I prefer some minimal test cases .

nhomar:
Monkeypatched for me is a :-1:  but it is a :+1: for the functionality.

pedrobaeza:
It's a WIP still, @nhomar. I will use another technique.

rafaelbn:
This is a WIP, @pedrobaeza are you going to continue it?

crm_claim_ref_smartbutton/README.rst
```diff
@@ -0,0 +1,64 @@
...
+  models is done on each view request, so this hits very little, but a bit,
+  the overall performance of the system.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/crm/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/crm/issues/new?body=module:%20crm_claim_ref_smartbutton%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
yajo, line 39: Please update with beautier newer template :blush:

crm_claim_ref_smartbutton/README.rst
```diff
@@ -0,0 +1,64 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
```
yajo, line 64: https

crm_claim_ref_smartbutton/i18n/crm_claim_ref_smartbutton.pot
```diff
@@ -0,0 +1,23 @@
+# Translation of Odoo Server.
```
yajo, line 1: Remove this file

crm_claim_ref_smartbutton/models/base_model.py
```diff
@@ -0,0 +1,111 @@
...
+
+# This should be monkey-patched always, or the computed field won't be
+# added to all models due to the module loading order
+models.BaseModel._add_magic_fields = _add_magic_fields_wrapper(
+    models.BaseModel._add_magic_fields)
+models.BaseModel.count_linked_claim_tech = count_linked_claim_tech
+
+
+class CrmClaimMonkeyPatcher(models.Model):
+    _name = 'crm.claim.monkey.patcher'
```
yajo, line 103: No need to create a model for this, you can simply add the register hook to any base model. I have seen it's common to use `ir.model`.

crm_claim_ref_smartbutton/models/base_model.py
```diff
@@ -0,0 +1,111 @@
...
+    _auto = False  # doesn't create table in database
+
+    def _register_hook(self, cr):
+        # Monkey patch functions - Only when the module is installed
+        models.BaseModel.action_open_linked_claims_tech = _open_related_claims
+        models.BaseModel.base_fields_view_get = base_fields_view_get
+        models.BaseModel.fields_view_get = fields_view_get_crm_claim
+        registered_dbs.append(cr.dbname)
```
yajo, line 111: Although monkey-patching is dirty by definition, this method patches all databases in the instance. It seems you use `registered_dbs` to keep track of that, but there's no need.

You can use [`_patch_method` and `_revert_method`](https://github.com/odoo/odoo/blob/8.0/openerp/models.py#L5210-L5247); those will take care of this with you. It's a bit complex but you can [inspire yourself by how base_field_validator does it](https://github.com/OCA/server-tools/blob/8.0/base_field_validator/models/ir_model.py#L34-L83).

crm_claim_ref_smartbutton/models/base_model.py
```diff
@@ -0,0 +1,111 @@
...
+# This should be monkey-patched always, or the computed field won't be
+# added to all models due to the module loading order
+models.BaseModel._add_magic_fields = _add_magic_fields_wrapper(
+    models.BaseModel._add_magic_fields)
+models.BaseModel.count_linked_claim_tech = count_linked_claim_tech
+
+
+class CrmClaimMonkeyPatcher(models.Model):
+    _name = 'crm.claim.monkey.patcher'
+    _auto = False  # doesn't create table in database
```
yajo, line 104: In case you still prefer to create a model just for this, you can simply set it as `AbstractModel`.

crm_claim_ref_smartbutton/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+from openerp.models import BaseModel
+from .models.base_model import registered_dbs
+from . import models
+
+
+def uninstall_hook(cr, registry):
+    # Undo monkey-patching to restore functionality
+    registered_dbs.remove(cr.dbname)
+    if not registered_dbs:
+        BaseModel.fields_view_get = BaseModel.base_fields_view_get
```
yajo, line 14: Of course, this shall change when you start using the right patching methods.

rafaelbn:
I guess you would like to finish this PR but will it be possible? Anyone can take this work and try to finish it?

PR #40, opened 2015-08-18
New Module Partner city autocomplete

Guillaumeleclercqbhc:
This module allows you to import the cities with zip and country. When you fill in a zip, the city will be automatically filtered based on this zip. When you select the city, the country will be add automatically. When the country is set,the related tax position for this country is add automatically.

2.0 : When you do install the module, this version adds the possibility to import, Zip codes and cities for Belgium and France. An option also allows you to import all other countries based on a CSV file.

I add the crm_blacklist module


pedrobaeza:
Hi, @Guillaumeleclercqbhc, thanks for the contribution, but OCA already has a module for this function: https://github.com/OCA/partner-contact/tree/8.0/base_location. Have you check it?

Guillaumeleclercqbhc:
Hi @pedrobaeza 
Ok,it's the same.  it's possible to delette the module partner city autocomplete ? But keep the black list module ?

pedrobaeza:
Well, I see that you should have some problems with the use of git, because there are 35 commits in this PR. Also the module should be in partner-contact repo, not here. So I recommend you to start with a fresh copy of partner-contact repo, copy in it the module, and make a single commit for the module. A renaming of the addon to _partner_blacklist_ should be also done.

I haven't discussed yet about the funcionality of the module itself, but I didn't go deep for now. I'll do it when the PR is issued.

hbrunn:
closed as of https://github.com/OCA/crm/pull/40#issuecomment-132208778

PR #42, opened 2015-08-28
[ADD] mass_mailing_partner

JavierIniesta:
This module links mass-mailing contacts with partners.


antespi:
@JavierIniesta Travis fails because of PEP8 issues, could you fix them? Thanks

mass_mailing_partner/README.rst
```diff
@@ -0,0 +1,63 @@
...
+
+
+Configuration
+=============
+
+At first install, all existing mass mailing contacts are matched against
+partners.
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
```
pedrobaeza, line 29: This goes in the Usage section

mass_mailing_partner/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+##############################################################################
+# License AGPL-3 - See LICENSE file on root folder for details
+##############################################################################
+
+from . import models
+from openerp import SUPERUSER_ID
+
+
+def _match_existing_contacts(cr, registry):
+    contact_model = registry['mail.mass_mailing.contact']
```
pedrobaeza: If you want, I have learnt the way to use new API here:

```
from openerp import api, SUPERUSER_ID

def ...(cr, registry):
    with api.Environment.manage():
        env = api.Environment(cr, SUPERUSER_ID, {})
        partner_model = env[res.partner]
        ...
```

mass_mailing_partner/__openerp__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+# License AGPL-3 - See LICENSE file on root folder for details
+##############################################################################
+
+{
+    "name": "Link partners with mass-mailing",
+    "version": "1.0",
```
pedrobaeza: Put version 8.0.1.0.0

mass_mailing_partner/views/res_partner_view.xml
```diff
@@ -0,0 +1,33 @@
...
+            <field name="domain">[('partner_id', '=', active_id)]</field>
+        </record>
+
+        <record model="ir.ui.view" id="view_partner_form_mass_mailing_contacts">
+            <field name="name">Partner Form with mailing contacts</field>
+            <field name="model">res.partner</field>
+            <field name="inherit_id" ref="base.view_partner_form"/>
+            <field name="arch" type="xml">
+                <div class="oe_right oe_button_box" position="inside">
+                    <button name="%(mass_mailing_partner.act_partner_mass_mailing_contacts)d"
```
pedrobaeza: You can use an existing act_window and save the declaration, translation and so on. You only need to put the same context you're using in the other act_window:

```
<button name="%(mass_mailing.action_view_mass_mailing_contacts)d"
        context="{'search_default_partner_id': active_id, 'default_partner_id': active_id}"
```

The only drawback is that the domain can be preserved, but I even prefer this option, because the filter is applied anyway, and you have the option to remove the filter and see all contacts without changing the screen.

dreispt:
I'm wondering if this module isn't best hosted at OCA/social?

pedrobaeza:
Well, in reality it is marketing, so crm/social are both side repositories where to host

dreispt:
There is a mass mailing feature proposed there: https://github.com/OCA/social/pull/9
It makes sense for both modules to be in the same repo.
"crm" is closer to "marketing" than "social", so maybe they should both go here.

pedrobaeza:
You're right that both are related. @antespi, can you check it?

rafaelbn:
@dreispt @pedrobaeza 
mass_mailing_partner is related with customers, how you work with them (crm)
mass_mailing_custom_unsubscribe is related with message functionality (social)
For ex. A social tools is mailchimp and https://github.com/OCA/social/pull/9 "is similar functionality" of mailchimp. A crm tool is SugarCRM and https://github.com/OCA/crm/pull/42 "is similar functionality" of SugarCRM.
So are totally different.

dreispt:
@rafaelbn OK

rafaelbn:
@dreispt just to be sure.. (eventhoug you say me OK) last question for you and @pedrobaeza 
I have answer today to your original thread "Renaming OCA/project-service to OCA/vertical-service". There are two things to take care about:
- Which project/repo - The name say something about which kind of modeles and funcionality you can find there
- Which name of module: the name of module is technicall, about what the module touch.
  This module touch in mass_mailing and in partners, must be "mass_mailing_partner" OR "partner_mass_mailing"?

I cannot read anything in https://odoo-community.org/ nor https://github.com/OCA who clarify me this doubts.
Thanks

antespi:
@JavierIniesta, I found thise errors on runbot:

## ERROR A:
1. Goes to Agrolait partner form
2. Click on 'Mailing list' smart buttton

**Current**: Appears all subscribers
**Expected**: None, Agrolait is not subscribed to any mailing list

## ERROR B:
1. Go to Marketing > Mailing lists > Contacts
2. Create
3. Fill partner column using Agrolait (email and name are filled automatic)
4. Save

```
Integrity Error

The operation cannot be completed, probably due to the following:
- deletion: you may be trying to delete a record while other records still reference it
- creation/update: a mandatory field is not correctly set

[object with reference: email - email]
```

mass_mailing_partner/views/mail_mass_mailing_contact_view.xml
```diff
@@ -0,0 +1,40 @@
...
+<data>
+
+<record model="ir.ui.view" id="view_mail_mass_mailing_contact_tree">
+    <field name="name">mail.mass_mailing.contact.tree.partner</field>
+    <field name="model">mail.mass_mailing.contact</field>
+    <field name="inherit_id" ref="mass_mailing.view_mail_mass_mailing_contact_tree"/>
+    <field name="arch" type="xml">
+        <field name="email" position="after">
+            <field name="partner_mandatory" invisible="1"/>
+            <field name="partner_id" attrs="{'required': [('partner_mandatory', '=', True)]}"/>
```
antespi:
```
attrs="{'required': [('partner_mandatory', '=', True)]}"
```

It's not neccesary, because if `partner_mandatory` is True. If no partner is filled then addon will search (or create) one with that email, check for duplicates and assigne partner field.

But if attrs defined, user can not create a contact filling only email. User is expecting that addon search for the partner (or create a new partner with that email if not exists)

antespi:
@JavierIniesta, please update spanish translation

JavierIniesta:
@antespi Code fixes and spanish translation updated

antespi:
:+1: Cool, thanks

rafaelbn:
Tested in runbot, searched for @antespi 's errors
:+1:  thanks!

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,87 @@
...
+    @api.one
+    @api.onchange('partner_id')
+    def _onchange_partner(self):
+        if self.partner_id:
+            self.name = self.partner_id.name
+            self.email = self.partner_id.email
+
+    @api.one
+    @api.constrains('partner_id')
+    def _check_partner_id(self):
```
pedrobaeza: This is better implemented with an unique SQL constraint.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,87 @@
...
+                    _('Partner already exists in this mailing list.'))
+
+    @api.model
+    def create(self, vals):
+        if not vals.get('partner_id'):
+            vals = self._check_partner(vals)
+        vals = self._check_name_email(vals)
+        return super(MailMassMailingContact, self).create(vals)
+
+    @api.one
```
pedrobaeza, line 33: This makes mass-write over this model very slow, as the operation is done record by record, and searching again and again for the same partner. You should rewrite it to support @api.multi and make the operations only one time. The only problem here is the mailing_list.partner_mandatory check, that in a multi-record environment, there can be some that has this flag marked and some not. You will have to check if `any(self.mapped('mailing_list.partner_mandatory')` and create the partner in that case.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,87 @@
...
+    def _prepare_partner(self, vals, mailing_list):
+        vals = {
+            'name': vals.get('name') or vals.get('email'),
+            'email': vals.get('email', False),
+        }
+        if mailing_list.partner_category:
+            vals['category_id'] = [(4, mailing_list.partner_category.id, 0)]
+        return vals
+
+    def _check_partner(self, vals):
```
pedrobaeza: Don't call it `_check`, because that's the usual name for @api.constrains methods.

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,87 @@
...
+            # Partner found
+            vals['partner_id'] = partners[0].id
+        elif mailing_list.partner_mandatory:
+            # Create partner
+            partner = m_partner.sudo().create(
+                self._prepare_partner(vals, mailing_list))
+            vals['partner_id'] = partner.id
+        return vals
+
+    def _check_name_email(self, vals):
```
pedrobaeza: Replace _check by another thing

mass_mailing_partner/models/res_partner.py
```diff
@@ -0,0 +1,37 @@
...
+    def _count_mass_mailing_contacts(self):
+        self.mass_mailing_contacts_count = len(self.mass_mailing_contacts)
+
+    @api.one
+    def write(self, vals):
+        old_name = self.name
+        old_email = self.email
+        res = super(ResPartner, self).write(vals)
+        if old_name != self.name or old_email != self.email:
+            if self.mass_mailing_contacts and not self.email:
```
pedrobaeza: This again doesn't take into account multi-record environments, raising an "Expected singleton" error if so. You should do `self.mapped('mass_mailing_contacts')` to merge all the contacts from all the affected records. But instead doing this here in the write, you should do it in a `@api.constrains('email')` method. You can use there the @api.one and simplify the logic.

mass_mailing_partner/tests/base.py
```diff
@@ -0,0 +1,37 @@
...
+from openerp.tests.common import TransactionCase
+
+
+class BaseCase(TransactionCase):
+    def setUp(self):
+        super(BaseCase, self).setUp()
+        m_data = self.registry("ir.model.data")
+        m_partner_category = self.env['res.partner.category']
+        m_mailing_list = self.env['mail.mass_mailing.list']
+        self.company_id = m_data.get_object_reference(
```
pedrobaeza: Why don't use `self.env.ref('base.main_company')`?

mass_mailing_partner/models/mail_mass_mailing_contact.py
antespi: This constrains can't be done with SQL because a list can be configured with no mandatory partner, so partner_id could be False in several records.

pedrobaeza: And that's exactly what an SQL constraint does: ignore null duplicities.

mass_mailing_partner/models/mail_mass_mailing_contact.py
JavierIniesta, line 33: I can't do this with @api.multi because vals may be different for each record.

mass_mailing_partner/models/res_partner.py
```diff
@@ -0,0 +1,38 @@
...
+    @api.one
+    @api.depends('mass_mailing_contacts')
+    def _count_mass_mailing_contacts(self):
+        self.mass_mailing_contacts_count = len(self.mass_mailing_contacts)
+
+    @api.multi
+    def write(self, vals):
+        res = super(ResPartner, self).write(vals)
+        if vals.get('name') or vals.get('email'):
+            for partner in self:
```
pedrobaeza: Do here:

```
mm_vals =  {}
if vals.get('name'):
    mm_vals['name'] = vals['name']
if vals.get('email'):
    mm_vals['email'] = vals['email']
self.mapped('mass_mailing_contacts').write(mm_vals)
```

mass_mailing_partner/__init__.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# See README.rst file on addon root folder for license details
+
+from . import models
+from openerp import api, SUPERUSER_ID
+
+
+def _match_existing_contacts(cr, registry):
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, context={})
```
hbrunn: `context` is not a keyword argument here

mass_mailing_partner/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+        contact_model = env['mail.mass_mailing.contact']
+        partner_model = env['res.partner']
+        contacts = contact_model.search([])
+        for contact in contacts:
+            if contact.email:
```
hbrunn, line 15: search for `[('email', '!=', False)]` above

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,75 @@
...
+    ]
+
+    @api.one
+    @api.onchange('partner_id')
+    def _onchange_partner(self):
+        if self.partner_id:
+            self.name = self.partner_id.name
+            self.email = self.partner_id.email
+
+    @api.model
```
hbrunn, line 25: add `@api.returns('self', lambda x: x.id)` to downgrade correctly for legacy code

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,75 @@
...
+            'email': vals.get('email', False),
+        }
+        if mailing_list.partner_category:
+            vals['category_id'] = [(4, mailing_list.partner_category.id, 0)]
+        return vals
+
+    def _set_partner(self, vals):
+        m_mailing = self.env['mail.mass_mailing.list']
+        m_partner = self.env['res.partner']
+        list_id = vals.get('list_id') or self.list_id.id
```
hbrunn: pass the fallback as second parameter to `dict.get`

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,75 @@
...
+        email = vals.get('email') or self.email
+        mailing_list = m_mailing.browse(list_id)
+        # Look for a partner with that email
+        email = email.strip()
```
hbrunn, line 59: is there a possibility that `email` is False?

mass_mailing_partner/models/mail_mass_mailing_contact.py
```diff
@@ -0,0 +1,75 @@
...
+        partners = m_partner.search([('email', '=ilike', email)], limit=1)
+        if partners:
+            # Partner found
+            vals['partner_id'] = partners[0].id
+        elif mailing_list.partner_mandatory:
+            # Create partner
+            partner = m_partner.sudo().create(
```
hbrunn, line 66: in which cases do we need `sudo` here?

mass_mailing_partner/models/res_partner.py
```diff
@@ -0,0 +1,41 @@
...
+        string='Mailing list number', compute='_count_mass_mailing_contacts',
+        store=True)
+
+    @api.one
+    @api.constrains('email')
+    def _check_email_mass_mailing_contacts(self):
+        if self.mass_mailing_contacts and not self.email:
+            raise ValidationError(
+                _("This partner '%s' is subscribed to one or more "
+                  "mailing lists." % self.name))
```
hbrunn: tell the user what the actual problem is (subscribed and no email)

mass_mailing_partner/models/mail_mass_mailing_contact.py
JavierIniesta, line 59: If create is called from another method, it is possible. I will check it and return `vals`. An error will raise validating the create fields.

mass_mailing_partner/models/mail_mass_mailing_contact.py
JavierIniesta, line 66: When user has permission to create mailing contacts but not partners.

JavierIniesta:
@hbrunn Changes done and questions answered. Thanks!

mass_mailing_partner/models/mail_mass_mailing_contact.py
hbrunn, line 66: I see. Not blocking for me, but in such cases, you might consider using https://github.com/OCA/server-tools/tree/8.0/base_suspend_security - this way, the partner will be created with the current user's uid, which looks a bit nicer than everything created by admin

hbrunn:
:+1:

PR #43, opened 2015-08-31
[ADD] Includes a way to classify claims adding the concept of type an…

yaniaular:
…d with this, it allows to configure claim stages depending on claim types. This module includes Customer, Supplier and Other claim types as default data and its own stages relation, and also makes few stages common for different claim types

### DIFF
--- a/crm_claim_type/README.rst
+++ b/crm_claim_type/README.rst
@@ -1,0 +1,69 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+CRM Claim Types
+===============
+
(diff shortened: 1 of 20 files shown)


pedrobaeza:
For me it seems now correct (although there are some nitpicking like the title style or the version numbering whose conventions have changed during the progress of this PR), but I don't want to block it more.

:+1:

crm_claim_type/README.rst
```diff
@@ -0,0 +1,69 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
```
lmignon, line 3: Make the module title a real title by adding '======' also above the title.

lmignon:
:+1: (Code review + functional tests)

nhomar:
two +1 then I merge.

PR #48, opened 2015-09-22
[IMP] More compatible with 7.0 crm_claim_rma

mistotebe:
To provide an upgrade path from crm_claim_rma from 7.0:
- The constraint needs to be relaxed to per company only
- `'number'` field used as fallback
- Provide `name_get()`

### DIFF
--- a/crm_claim_code/__openerp__.py
+++ b/crm_claim_code/__openerp__.py
@@ -22,11 +22,12 @@
 
 {
     "name": "Sequential Code for Claims",
-    "version": "1.0",
+    "version": "8.0.1.1.0",
     "category": "Customer Relationship Management",
     "author": "OdooMRP team, "
               "AvanzOSC, "
               "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
+              "credativ ltd., "
               "Odoo Community Association (OCA)",
     "website": "http://www.avanzosc.es",
     "license": "AGPL-3",
--- a/crm_claim_code/models/crm_claim.py
+++ b/crm_claim_code/models/crm_claim.py
@@ -32,10 +32,12 @@ class CrmClaim(models.Model):
         return super(CrmClaim, self).copy(default)
 
     @api.multi
-    @api.depends('code', 'name')
+    @api.depends('code')
     def name_get(self):
+        orig_names = dict(super(CrmClaim, self).name_get())
+
         res = []
         for claim in self:
-            name = "[{0.code}] {0.name}".format(claim)
+            name = "[{}] {}".format(claim.code, orig_names[claim.id])
             res.append((claim.id, name))
         return res
--- a/crm_claim_code/models/crm_claim.py
+++ b/crm_claim_code/models/crm_claim.py
@@ -10,11 +10,12 @@ class CrmClaim(models.Model):
     _inherit = "crm.claim"
 
     code = fields.Char(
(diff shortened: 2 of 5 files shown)


crm_claim_code/models/crm_claim.py
```diff
@@ -29,3 +30,12 @@ def copy(self, default=None):
             default = {}
         default['code'] = self.env['ir.sequence'].get('crm.claim')
         return super(CrmClaim, self).copy(default)
+
+    @api.multi
+    @api.depends('code', 'name')
```
pedrobaeza: @api.depends is for computed fields, not for this one.

You should also call super and prefix the code instead hardcoding it for allowing other personalizations.

crm_claim_code/__openerp__.py
```diff
@@ -22,7 +22,7 @@
 
 {
     "name": "Sequential Code for Claims",
-    "version": "1.0",
+    "version": "1.1",
```
pedrobaeza: Use new version numbering scheme 8.0.1.1.0

mistotebe:
Re `@api.depends`: this is how it works in v8, it is a field display_name with function name_get(). Quite a few v9 modules use it, event.event in the event module is an example that uses it in v8. See odoo/odoo@cbe2dbb6 for more information.

crm_claim_code/models/crm_claim.py
```diff
@@ -29,3 +30,14 @@ def copy(self, default=None):
             default = {}
         default['code'] = self.env['ir.sequence'].get('crm.claim')
         return super(CrmClaim, self).copy(default)
+
+    @api.multi
+    @api.depends('code')
```
pedrobaeza, line 35: My comment has been lost, but again I insist depends is not needed. It's not a computed field, it's a method.

mistotebe, line 35: Again, since v8 it is a computed field called display_name with function name_get().

pedrobaeza, line 35: No, it isn't a new API computed field, which is the one that needs depends. If it was a new computed field, it won't return a value, but assign it. Old style computed fields require triggers.

mistotebe, line 35: Surely Fabien would know what he's doing :)
https://github.com/odoo/odoo/commit/b254dad80

mistotebe, line 35: I'm not saying it's 100% needed, just that this is how the new API works and the decorator uses it exactly as intended (and unlike the v7/v8 compatibility layer which actually slows things down, this will not have negative perf. impact, the opposite if I understand what the next steps in this area will be). As such, I'm not sure why it's bothering you.

pedrobaeza, line 35: Wow, I'm totally shocked now by this use that I haven't found, but it seems that both uses (depends and triggers) are interoperable in new and old API fields, so cool, let's keep it. Thanks for opening my eyes in this.

mistotebe, line 35: No worries! I'm still finding cool new use cases with the new API even now. Makes some things so much easier to express.

pedrobaeza:
Can you please rebase and update your code accordingly?

rafaelbn:
This is one year old PR without anwwer, closing

PR #52, opened 2015-10-06
Lead vat

yajo:
Relaunching part of https://github.com/Antiun/antiun-odoo-addons/pull/11.

Add VAT field to leads.

It will fully work after https://github.com/odoo/odoo/pull/8899 is merged, but I think it can be merged anyway.

cc @rafaelbn.


yajo:
Now NOT depending on https://github.com/odoo/odoo/pull/8899.

rafaelbn:
Tested in runbot :+1:

pedrobaeza:
Please check Travis

yajo:
Local tests work, [seems like new_api = None in Travis](https://travis-ci.org/OCA/crm/jobs/85146073#L285) for some reason. I did leap of faith with last commit. Let's see...

yajo:
Travis hates me. Now [You try to install module 'crm_lead_address_street3' that depends on module 'partner_address_street3'.\nBut the latter module is not available in your system](https://travis-ci.org/OCA/crm/jobs/85753133#L229). Nothing to do with this PR.

pedrobaeza:
I have fixed the problem in https://github.com/OCA/crm/commit/d4c13487a16d10d342ecf3e371225153767e9dea. Please rebase and you'll get a green branch.

yajo:
https://github.com/OCA/crm/pull/52#issuecomment-148742486 still happening.

crm_lead_vat/models.py
```diff
@@ -0,0 +1,31 @@
...
+    vat = fields.Char(
+        "TIN",
+        help="Tax Identification Number. The first 2 characters are the "
+             "country code.")
+
+    @api.model
+    def _lead_create_contact(self, lead, name, is_company, parent_id=False):
+        """Add VAT to partner."""
+        return (super(Lead, self.with_context(default_vat=lead.vat))
+                ._lead_create_contact(lead, name, is_company, parent_id))
```
yajo, line 20: [This is the line that fails in Travis](https://travis-ci.org/OCA/crm/jobs/86186745#L308), but as you can see this [is almost a copy of this other](https://github.com/OCA/l10n-spain/pull/224/files#diff-e5d75ea06e4728330dc5997a57171873R18), which is in another OCA repo and Travis is OK. Local tests work too.

osvalr, line 20: Hi @Yajo,

Did you try with **`getattr`** as the following

``` python
return getattr(super(Lead, self.with_context(default_vat=lead.vat)),'_lead_create_contact')(lead, name, is_company, parent_id)
```

yajo, line 20: It's the same.

crm_lead_vat/tests/test_lead.py
```diff
@@ -0,0 +1,29 @@
...
+    def test_transfered_values(self):
+        """Field gets transfered when creating partner."""
+        self.lead.vat = self.test_field
+        self.lead.handle_partner_assignation()
+        self.assertEqual(self.lead.partner_id.vat, self.test_field)
+
+    def test_onchange_partner_id(self):
+        """Lead gets VAT from partner when linked to it."""
+        self.partner.vat = self.test_field
+        self.lead.partner_id = self.partner
```
pedrobaeza, line 27: This line is not needed

pedrobaeza:
Maybe some modules are coliisioning. Try to add to your PR an specific .travis.yml for testing only your module with INCLUDE option. See https://github.com/OCA/purchase-workflow/blob/8.0/.travis.yml#L21

crm_lead_vat/tests/test_lead.py
yajo, line 27: No, but it is there for imitating the UI workflow.

yajo:
I'm trying another idea, let's see...

crm_lead_vat/tests/test_lead.py
yajo, line 27: Sorry, this is actually needed.

pedrobaeza, line 27: For what? Is an old API onchange. All the values needed must be taken from the arguments, not from the record values (that can be outdated).

yajo, line 27: I was actually doing that, and it works fine.

yajo:
Finally! Thanks @pedrobaeza, that seems to work (not my previous idea). No idea of what is the conflicting module though.

pedrobaeza:
I think the problem is here: https://github.com/OCA/crm/blob/9.0/crm_lead_address_street3/model/crm_lead.py#L30, because there's no corresponding @api.v8 method. Can you please check it and fix it?

pedrobaeza:
:+1:

rafaelbn:
@Yajo please fix this module wiith a new PR, same problem as https://github.com/OCA/crm/pull/53#issuecomment-152221135

yajo:
See https://github.com/OCA/crm/pull/62

PR #56, opened 2015-10-08
[ADD] marketing_security_group addon

antespi:
This module add a this security features:
- ACL for allowing marketing user group to remove mass.mailing objects
- Security rule to allows marketing user to delete only his not sent mail.mass_mailing objects
- Security rule to allows marketing manager to delete all not sent mail.mass_mailing objects

For sent mail.mass_mailing objects, only an user in Administration / Settings group
could remove them, like Odoo standard defines


pedrobaeza:
You don't need to manually maintain README.md module list. It's done automatically by an script. About the module name, it's related to mass_mailing, so maybe you should rename it and go to OCA/social instead here

antespi:
Repo README changes discarted, squash and rebase.

@pedrobaeza, I think this addon belongs to this repository. It inherits from marketing to allow marketing groups to do an extra operation over mass_mailing object. It's not related with Odoo messaging system. This case is similar to `mass_mailing_partner` addon, that also inherit from mass_mailing.

pedrobaeza:
OCA/social is not for Odoo messaging system. It's for interaction with others, like mass_mailing is.

rafaelbn:
@pedrobaeza @oca-clabot I think @antespi has signed CLA some months ago please review. Thanks
@antespi I think we should move this addon to OCA/social . We don't have OCA/Marketing because OCA/social. Please move it. Thanks

JavierIniesta:
Tested in local environment :+1:

antespi:
Moved to social repo, with all fixes.

PR #55, opened 2015-10-06
Add module crm_lead_supplier.

yajo:
Relaunching last part of Antiun/antiun-odoo-addons#11.

Add customer/supplier checkbox fields to leads.

It will fully work after odoo/odoo#8899 is merged, but I think it can be merged anyway.

cc @rafaelbn.

PR #54, opened 2015-10-06
Add module crm_lead_invoice_address.

yajo:
Relaunching part of Antiun/antiun-odoo-addons#11.

Add invoice address fields to leads.

It will fully work after odoo/odoo#8899 is merged, but I think it can be merged anyway.

cc @rafaelbn.

PR #53, opened 2015-10-06
Add module crm_lead_website.

yajo:
Relaunching part of Antiun/antiun-odoo-addons#11.

Add website field to leads.

It will fully work after odoo/odoo#8899 is merged, but I think it can be merged anyway.

cc @rafaelbn.

PR #61, opened 2015-10-23
Update crm_lead_address_street3 to fix Travis.

yajo:
Fix Travis for https://github.com/OCA/crm/pull/52, https://github.com/OCA/crm/pull/53, https://github.com/OCA/crm/pull/54 and https://github.com/OCA/crm/pull/55.

cc @pedrobaeza @rafaelbn

### DIFF
--- a/crm_lead_address_street3/model/crm_lead.py
+++ b/crm_lead_address_street3/model/crm_lead.py
@@ -21,7 +21,7 @@
 from openerp import models, fields, api
 
 
-class Lead(models.Model):
+class CrmLead(models.Model):
     """Add third field in lead address"""
     _inherit = "crm.lead"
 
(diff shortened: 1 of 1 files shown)


yajo:
I hope you can merge this ASAP to continue with all other PR. Anyway all are rebased on this now.

crm_lead_address_street3/model/crm_lead.py
```diff
@@ -21,31 +21,21 @@
 from openerp import models, fields, api
 
 
-class crm_lead(models.Model):
+class Lead(models.Model):
```
pedrobaeza: Call if `CrmLead`

pedrobaeza:
@StefanRijnhart, there was any reason to put @api.v7?

StefanRijnhart:
We can ask the author @rdeheele but I don't think so. :+1: for now. What problem does @api.v7 cause in Travis? I can't check Travis in the linked PRs, mabye because they have been rebased.

yajo:
See all linked PR.

yajo:
Mostly https://github.com/OCA/crm/pull/52#discussion_r42514858.

StefanRijnhart:
Thanks for the link, and for fixing this! :+1:

rdeheele:
@pedrobaeza @StefanRijnhart i don't remember exactly, there are cr and uid parameters, maybe it was easier to use api.v7 to inherit a core class itself in a v7 api style, 8 months ago.

yajo:
Really api.v7 had not much sense, since it is used only in conjuntion with api.v8 to provide different implementations for the same method depending on the calling method. Since there was no api.v8, all calls from v8 were failing.

rdeheele:
ok :+1:

PR #62, opened 2015-10-29
Add field to opportunities form too.

yajo:
CC @rafaelbn @carlos-incaser @sergio-incaser @pedrobaeza.


carlosdauden:
:+1:

janetrp:
question: TIN in spanish should be CIF? Tested in Runbot.
VAT field (TIN) appears also in opportunities
:+1:

yajo:
Weird, Odoo does not generate the correct .pot file and I cannot translate! :fearful: What to do?

rafaelbn:
:+1:

PR #76, opened 2016-02-18
[8.0][crm_deduplicate_acl] Deduplicate permission groups

yajo:
This module allows any user to get special permissions to be able to
deduplicate contacts, instead of just giving them to a sale manager with
settings permissions.

The new permissions allow the user to deduplicate:
- _Manually_.
- _Automatically_ (dangerous).
- _Without restrictions_ (more dangerous).

Beware what you do! :hushed:

@rafaelbn


crm_deduplicate_acl/README.rst
```diff
@@ -0,0 +1,95 @@
...
+
+#. Go to *Settings > Users > Users*.
+#. Choose a user.
+#. Choose the desired permission level in *Appplication > Deduplicate
+   Contacts*:
+
+   - *Manually* allows him to do the manual deduplication process.
+   - *Automatically* allows him to do the automatic deduplication process.
+
+     .. warning::
```
pedrobaeza, line 26: This is not well rendered on RST

crm_deduplicate_acl/wizards/partner_merge.py
```diff
@@ -0,0 +1,25 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import SUPERUSER_ID, api, models
+
+
+class BasePartnerMergeAutomaticWizard(models.TransientModel):
+    _inherit = "base.partner.merge.automatic.wizard"
+
+    @api.cr_uid_context
+    def _merge(self, cr, uid, partner_ids, dst_partner=None, context=None):
```
pedrobaeza, line 12: Why not using new API (@api.model)?

pedrobaeza:
Some little tests that exercises the different levels would be welcome.

rafaelbn:
Please add permission to this group (option Manually) for merging partners with different e-mails



Thanks

crm_deduplicate_acl/README.rst
yajo, line 26: I hope they fix the theme some day. In the mean time it writes "Warning", should be enough don't you think?

captura de pantalla de 2016-02-19 16-43-18

crm_deduplicate_acl/wizards/partner_merge.py
yajo, line 12: See known issues.

yajo:
@rafaelbn You need the unrestricted access for that.

rafaelbn:
But in the case a user who is responsible to maintain clean the database have to merge partners with different email but you still want to force manualy, you can't. That's why

rafaelbn:
Please review my use case and runbot, thanks!

yajo:
@rafaelbn Fixed https://github.com/OCA/crm/pull/76#issuecomment-186311807

yajo:
What happens to runbot?

rafaelbn:
Tested and running in customer's production enviroment :+1:

sergio-teruel:
Tested in local environment.... :+1:

crm_deduplicate_acl/wizards/partner_merge.py
pedrobaeza, line 12: You can implement this with `@api.v7` and `@api.v8` methods

yajo, line 12: I don't see the need to duplicate efforts, api takes care of morphing calls from v8.

pedrobaeza, line 12: It's not to duplicate efforts. See https://github.com/odoo/odoo/blob/8.0/addons/account/account_invoice.py#L1209-L1213

yajo, line 12: [The problem is the missing model_cr decorator, that has already landed in master](https://github.com/odoo/odoo/blob/master/openerp/api.py#L435). For v8 I could do that, but it's not called from v8 methods anyway.

pedrobaeza, line 12: The trick is to make a v7 wrapper, and call to the v8 method from that wrapper, and you do the hard coding all inside the v8 method, with new API. Anyway, it's already merged in the repository by the impatient @rafaelbn :stuck_out_tongue_winking_eye:

Issue #83, opened 2016-03-10
[8.0][lettermgmt] Module simplification


ivantodorovich:
I intend to use lettermgmt in our own company, but I found the module to be too much complicated for our current needs.

I'm wondering if @hbrunn or @bwrsandman can throw some clarification on a few things. I intend to improve the readme once I get a full sense of the module, and to further develop it.
### Letter States / Workflow

_Draft -> Created -> Validated -> Sent_

What's the purpose of this workflow? Why having both Created and Validated?
### Fields clarification / Intention

What's the purpose of these fields? Can you provide examples?

**parent:** How can a letter be parent/child of another letter?
**thread / child_ids:** I take it it's related to parent.. but can't seem to find an use case.

**type**: What's the type of letter? Is it the courier / postal service used? (IE: "48hs delivery") Could it be the kind of envelope used?
**class:** ¿? Is this supposed to be like internal categories of letters? (IE: "Invoice", "Receipt".. )
**channel:** Courier / postal service? Internal office? (IE: "FedEx" or maybe an internal office: "Front Desk")

**rassigment_ids**: Purpose/Example use case?

**Letter Folders**: Is it for big companies with a lot of letter transit locations?

max3903:
@bwrsandman worked on it for a large organization with a internal courrier service in charge of receiving and shipping letters and parcels internally and externally.

# Workflow

You can have people creating the letter and people verifying that the information entered in the system is the one on the letters/parcels.

# Fields

parent: You can have one incoming shipment spread out among many letters/boxes.

This is as much as I can remember about it. Maybe @plamarche have more information.

ivantodorovich:
Thanks for the clarification! :+1: 

I'll wait for the others' response

hbrunn:
I just ported what existed, and actually wrote a module that dumbs this down be removing all fields you ask about save for `class`, `channel` and `type`. About those, have a look at the demo data in https://github.com/OCA/crm/blob/8.0/lettermgmt/data/letter_demo.xml to confirm you're mostly right about their use.

ivantodorovich:
Thank you @hbrunn !

That module you wrote, the one that dumbs this down, is it public?

I'd like to hear your thought and @bwrsandman's about splitting this module in a few: a simplified base + specific extensions. I'll gladly make that PR, but only with your approval.

hbrunn:
[...]

PR #87, opened 2016-03-17
[8.0] Refactor lettermgmt

ivantodorovich:
As discussed in https://github.com/OCA/crm/issues/83

Feel free to comment and make suggestions as I work.

TODO:
- [x] ~~Migrate field class to class_id~~ _migrated to category_ids._
- [x] ~~Add seccurity group to show/hide reassignments~~
- [x] ~~Add seccurity group to show/hide partner / child~~
- [x] ~~Increment sequence only after the record is saved, not before.~~
- [x] ~~Simplify and improve state workflow.~~
- [ ] ~~Support current "advanced" workflow in a separate module.~~ _wont support it in this pr_
- [x] ~~Add default filter 'Pending' and search fields to list views.~~
- [x] ~~Add tests~~


hbrunn:
another great thing to add would be tests...

ivantodorovich:
@hbrunn 
Please take a look when you have the time.

**Functional changes:**
- Field `class` is now a Many2many_tags called `category_ids`
- Field `date` is now field.Date (instead of Datetime).
- Tree View improvements
  - Added colours
  - Added a default filter 'Pending' to show letters that are not received yet (either by us or by the external partner)
- The workflow for `in` and `out` letters is now the same. 
  - This way we can track letters sent until we get confrimation of reception and manually set it to 'received'
  - `snd_rec_date` is replaced by fields `snd_date` and `rec_date`
  - `snd_date` is the date the letter was sent either by us (`out`) or by the external sender (`in`)
  - `rec_date` is the date the letter was received by us (`in`) or the date the recipient confirmed (`out`)
- Simplified workflow
  - Removed `created` and `validated` states.
  - Current workflow is: 
    - `Draft -> Sent -> Received / Rec. Returned / Rec. Bad`
    - `Cancelled -> Back to Draft`
- Groups are added to show/hide some fields for extended usage
  - group to show/hide parent and childs ("thread")
  - group to show/hide reassignments
- Sender and Recipient are now defaulted to current company, depending on the letter move (`in` or `out`)
- It no longer add subscribers when changing the state. To my understanding, it was annoying.
- Some other changes..

ivantodorovich:
I'm need to make the module to extend the workflow. It will re-implement the 'created' and 'validated' states.

I still don't completely understand the use case, though.. I mean, it didn't have security groups, so any user was able to create or validate any letter. If it was only to track the letter prep states before sending, the same could be achived by just using internal notes or messages.

Even so, it seems to be a very company-specific case. I think it's very possible no one will use it anymore (and ever). 

Should I do it the same? I don't mind.. I just don't like the idea of useless modules.

ivantodorovich:
@max3903 I'd like to hear your thoughts too

hbrunn:
what's the state of this PR? Can we review?

hbrunn:
(asking because I'm in the process of adding a 'done' state and support for needaction, and it would make sense to get your refactoring merged first of course)

lettermgmt/migrations/8.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(cr, version):
+    if version is None:
+        return
+
+    # Migrate 'type' to 'type_id'
+    cr.execute("""ALTER TABLE res_letter
+                  RENAME COLUMN type to type_id""")
```
hbrunn: you should borrow openupgrade's code for renaming a model here: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L328 to get links from other models (mostly) right after you renamed `letter.class`

lettermgmt/migrations/8.0.2.0/post-migration.py
```diff
@@ -0,0 +1,30 @@
...
+def migrate(cr, version):
+    if version is None:
+        return
+
+    # Migrate states ['created', 'validated'] back to 'draft'
+    cr.execute("""UPDATE res_letter SET state = 'draft'
+                  WHERE state IN ('created', 'validated')""")
+
+    # Migrate 'letter.class' to 'letter.category'
+    cr.execute("""INSERT INTO letter_category SELECT * FROM letter_class""")
```
hbrunn: better to this in pre-migrate, see below

ivantodorovich:
> @hbrunn 
> what's the state of this PR? Can we review?
> (asking because I'm in the process of adding a 'done' state and support for needaction, and it would make sense to get your refactoring merged first of course)

Well, I kinda stop working on it because of the low attention it got.
But I believe this PR was almost ready, except for the 'advanced workflow' support, which is removed and it should be implemented in a separate module because of its rare usage.. but maybe no one will ever use it.

I'll take a look at openupgrade's code, as you suggested, and make the changes later this week.

hbrunn:
that's probably because of the work in progress label - I changed this to needs review to invite other reviewers

lettermgmt/models/res_letter.py
```diff
@@ -1,165 +1,200 @@
...
-        """Put the state of the letter into Validated"""
-        for letter in self.browse(cr, uid, ids, context=context):
-            self.write(
-                cr, uid, [letter.id], {'state': 'validated'}, context=context)
+    @api.multi
+    def action_cancel_draft(self):
+        """ Go from cancelled state to draf state """
+        self.write({'state': 'draft'})
+        self.delete_workflow()
+        self.create_workflow()
```
hbrunn: this model doesn't define a workflow (in the sense of odoo's workflows), so why do you need this?

lettermgmt/models/res_letter.py
```diff
@@ -1,165 +1,200 @@
...
-                },
-                context=context)
-            if letter.sender_partner_id:
-                letter.message_subscribe([letter.sender_partner_id.id])
+    @api.multi
+    def action_send(self):
+        """ Put the state of the letter into sent """
+        self.write({
+            'state': 'sent',
+            'snd_date': self.snd_date or fields.Date.today()
```
hbrunn, line 169: this is going to break for n > 1. You really need the loop

lettermgmt/models/res_letter.py
```diff
@@ -1,165 +1,200 @@
...
-        """Put the state of the letter into Received but Returned"""
-        for letter in self.browse(cr, uid, ids, context=context):
-            self.write(
-                cr, uid, [letter.id], {'state': 'rec_ret'}, context=context)
+    @api.multi
+    def action_received(self):
+        """ Put the state of the letter into Received """
+        self.write({
+            'state': 'rec',
+            'rec_date': self.rec_date or fields.Date.today()
```
hbrunn, line 178: same here

lettermgmt/models/res_letter.py
```diff
@@ -1,165 +1,200 @@
...
-        """Put the state of the letter into Received but Damaged"""
-        for letter in self.browse(cr, uid, ids, context=context):
-            self.write(
-                cr, uid, [letter.id], {'state': 'rec_bad'}, context=context)
+    @api.multi
+    def action_rec_ret(self):
+        """ Put the state of the letter into Received but Returned """
+        self.write({
+            'state': 'rec_ret',
+            'rec_date': self.rec_date or fields.Date.today()
```
hbrunn, line 187: and here

lettermgmt/models/res_letter.py
```diff
@@ -1,165 +1,200 @@
...
-        """Put the state of the letter into draft"""
-        for letter in self.browse(cr, uid, ids, context=context):
-            self.write(
-                cr, uid, [letter.id], {'state': 'draft'}, context=context)
+    @api.multi
+    def action_rec_bad(self):
+        """ Put the state of the letter into Received but Damaged """
+        self.write({
+            'state': 'rec_bad',
+            'rec_date': self.rec_date or fields.Date.today()
```
hbrunn, line 196: same

lettermgmt/views/letter_category_view.xml
```diff
@@ -0,0 +1,62 @@
...
+     © 2009 Parthiv Patel, Tech Receptives
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<openerp>
+  <data>
+
+    <!-- Views -->
+
+    <record model="ir.ui.view" id="letter_category_tree_view">
+      <field name="name">letter.category.tree</field>
```
hbrunn: this is nearly what odoo autogenerates as name when you leave out the field, so you could do just that

lettermgmt/models/res_letter.py
ivantodorovich: At the time of this PR, I was just starting to learn odoo. I guess I've just copied this from another module's code, and it doesn't apply here at all.

lettermgmt/models/res_letter.py
ivantodorovich: Yeah.. sorry about that. I should've reviewed past me myself.
I'll fix all of this later this week, probably Friday.

lettermgmt/migrations/8.0.2.0/pre-migration.py
ivantodorovich: Done

lettermgmt/migrations/8.0.2.0/post-migration.py
ivantodorovich: I couldn't do it in pre-migrate. I tried a few approaches and it kept giving me errors.

ivantodorovich:
I need help with migration. I'm not happy with how it is right now. I'm not sure if it works properly either.

hbrunn:
@ivantodorovich just from looking at the code, the migration seems okay. What is it you're not happy about?

ivantodorovich:
This right here:

`cr.execute("""INSERT INTO letter_category SELECT * FROM letter_class""")`

openupgrade renames the table instead of copying all the records from the old table into the new table.
I tried to do it that way, in pre-migration, but odoo won't accept it for some reason I couldn't figure out.

Anyway, considering this module's popularity, I guess no one will ever care about that, as long as it works.

hbrunn:
what error message do you get here? And did you prepare your test database accordingly?

rafaelbn:
Do you need any functional help here @ivantodorovich @hbrunn ?

ivantodorovich:
I'm sorry for not responding earlier. 
I haven't been able to make up the time this week.

I'll test again the migration code on a clean database and see if I can improve it. 
Any help reviewing the migration would be great.

pedrobaeza:
This PR is hard to review, but I can trust @hbrunn if he says this is OK

pedrobaeza:
@ivantodorovich, can you squash a bit the commits?

@hbrunn, any comments?

ivantodorovich:
Done. Squashed into a single commit.

hbrunn:
my comment is let's merge as soon as @ivantodorovich has fixed the last travis nags: https://travis-ci.org/OCA/crm/jobs/168795594#L314

hbrunn:
better link: https://travis-ci.org/OCA/crm/jobs/168795594#L307-L323

ivantodorovich:
Done, sorry about that!

pedrobaeza:
Thanks for your efforts!

PR #89, opened 2016-04-22
[FIX] Exception when scheduling a phonecall from another

antespi:
This PR fix a bug when scheduling a phonecall from another. Steps to reproduce:
1. Install addon 'crm_phonecall_summary_predefined'
2. Go to Sales > Logged phones and clean filters
3. Click at Phone icon (icons on the right in each call row) to 'Schedule Other Call'
4. Press 'Schedule Call' button

An IntegrityError is raised:

```
Integrity Error

The operation cannot be completed, probably due to the following:
- deletion: you may be trying to delete a record while other records still reference it
- creation/update: a mandatory field is not correctly set

[object with reference: summary_id - summary.id]
```

This PR adds summary_id field to `crm.phonecall2phonecall` model and view in order to create a new call with that field assigned and avoid integrity error.

This PR depends on these ones:
- https://github.com/OCA/OCB/pull/482
- https://github.com/odoo/odoo/pull/11791

### DIFF
--- a/crm_phonecall_summary_predefined/models/crm_phonecall.py
+++ b/crm_phonecall_summary_predefined/models/crm_phonecall.py
@@ -54,7 +54,6 @@ class CRMPhonecall(models.Model):
         res = super(CRMPhonecall, self)._prepare_another_phonecall_vals(
             call, schedule_time, call_summary, user_id=user_id,
             section_id=section_id, categ_id=categ_id)
-        summary_id = (self.env.context and
-                      self.env.context.get('summary_id', False) or False)
-        res['summary_id'] = summary_id
+        res['summary_id'] = self.env.context.get(
+            'summary_id', res.get('summary_id', False))
         return res
--- a/crm_phonecall_summary_predefined/models/crm_phonecall.py
+++ b/crm_phonecall_summary_predefined/models/crm_phonecall.py
@@ -48,7 +48,7 @@ class CRMPhonecall(models.Model):
                 s.summary_id = summary.search([("name", "=", s.name)])
 
     @api.model
-    def _prepare_another_phonecall(
+    def _prepare_another_phonecall_vals(
             self, call, schedule_time, call_summary, user_id=False,
             section_id=False, categ_id=False):
         res = super(CRMPhonecall, self)._prepare_another_phonecall(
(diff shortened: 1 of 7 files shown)


crm_phonecall_summary_predefined/models/crm_phonecall.py
```diff
@@ -47,16 +47,14 @@ def _init_summary_id(self):
...
-        "Phonecalls",
-        help="Phonecalls with this summary.")
+    @api.model
+    def _prepare_another_phonecall_vals(
+            self, call, schedule_time, call_summary, user_id=False,
+            section_id=False, categ_id=False):
+        res = super(CRMPhonecall, self)._prepare_another_phonecall_vals(
+            call, schedule_time, call_summary, user_id=user_id,
+            section_id=section_id, categ_id=categ_id)
+        summary_id = (self.env.context and
```
pedrobaeza: You should do `res['summary_id'] = self.env.context.get('summary_id', res.get('summary_id', False))`.

crm_phonecall_summary_predefined/views/crm_phonecall_view.xml
```diff
@@ -33,6 +33,20 @@
         </field>
     </record>
 
+    <record id="phonecall_to_phonecall_view" model="ir.ui.view">
+        <field name="name">Replace name by summary_id</field>
+        <field name="model">crm.phonecall2phonecall</field>
+        <field name="inherit_id" ref="crm.phonecall_to_phonecall_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='name']" position="attributes">
```
pedrobaeza, line 41: Shouldn't you allow a free summary?

antespi:
Until https://github.com/OCA/OCB/pull/482 or https://github.com/odoo/odoo/pull/11791 merged, travis and runbot will be red

pedrobaeza:
This can't be merged on OCB, so the only chance is Odoo.

antespi:
Well, I hope Odoo HQ takes my PR into account. If not, we can see another approach to solve this bug.

crm_phonecall_summary_predefined/views/crm_phonecall_view.xml
antespi, line 41: This addon is for forcing to select for a defined list

pedrobaeza:
Can you add a test for the new behavior?

antespi:
Odoo is ignoring us and OCB fix is not an option, so I have to change strategy. Then when I have a new PR for this (without no core changes) I'll add test for it.

rafaelbn:
Hi @antespi @pedrobaeza , none of the dependencies of this PR are going to be merged. So Please I think it's better to close this. Anyway we are migrating this problem to v9, so better we must find a way in v9.

antespi:
I'm closing this because requirements are not going to be merged.

PR #90, opened 2016-04-26
crm_claim_code: migration v9

cyrilgdn:


crm_claim_code/__openerp__.py
```diff
@@ -1,28 +1,10 @@
...
+# © 2015 Serv. Tec. Avanzados - Pedro M. Baeza (http://www.serviciosbaeza.com)
+# © 2015 AvanzOsc (http://www.avanzosc.es)
 
 {
     "name": "Sequential Code for Claims",
-    "version": "8.0.1.0.0",
+    "version": "9.0.1.0.0",
     "category": "Customer Relationship Management",
     "author": "OdooMRP team, "
               "AvanzOSC, "
```
pedrobaeza: Can you please switch "Serv. Tecnol. Avanzados - Pedro M. Baeza" to my new company "Tecnativa" (it's even shorter!)?

pedrobaeza:
What I don't get is why the warning on runbot about using .get() when it has been removed. Any idea?

crm_claim_code/__openerp__.py
```diff
@@ -1,41 +1,24 @@
...
-              "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
+              "Tecnativa - Pedro M. Baeza, "
```
pedrobaeza: Please, only "Tecnativa", not "Tecnativa - Pedro M. Baeza"

cyrilgdn:
@pedrobaeza I forgot one ir_sequence.get in post_init hook, fixed!

crm_claim_code/__init__.py
```diff
@@ -15,11 +16,12 @@ def create_code_equal_to_id(cr):
 
 
 def assign_old_sequences(cr, registry):
-    claim_obj = registry['crm.claim']
-    sequence_obj = registry['ir.sequence']
-    claim_ids = claim_obj.search(cr, SUPERUSER_ID, [], order="id")
+    env = Environment(cr, SUPERUSER_ID, {})
```
pedrobaeza: I always make this with:

```
with api.Environment.manage():
    env = api.Environment(cr, SUPERUSER_ID, {})
```

I don't know if this has any side effect.

guewen: Yes I think we are supposed to do it.
IIRC, if we are already inside a `with api.Environment.manage()` which is the case most of the time (we are when we come from a rpc request), it has no effect, but it is safer to use `api.Environment.manage()` to handle cases where we would not already be in a `manage()` context.

crm_claim_code/models/crm_claim.py
```diff
@@ -20,12 +20,16 @@ class CrmClaim(models.Model):
...
-    @api.one
+    @api.multi
     def copy(self, default=None):
         if default is None:
             default = {}
-        default['code'] = self.env['ir.sequence'].get('crm.claim')
-        return super(CrmClaim, self).copy(default)
+        results = self.env['crm.claim']
+        for record in self:
+            default['code'] = self.env['ir.sequence'].next_by_code('crm.claim')
```
guewen, line 32: Better to use:

``` python
default.setdefault('code', self.env['ir.sequence'].next_by_code('crm.claim'))
```

So if the caller had explicitly set a default for `code`, it would be kept.

crm_claim_code/models/crm_claim.py
```diff
@@ -20,12 +20,16 @@ class CrmClaim(models.Model):
...
         return super(CrmClaim, self).create(vals)
 
-    @api.one
+    @api.multi
     def copy(self, default=None):
         if default is None:
             default = {}
-        default['code'] = self.env['ir.sequence'].get('crm.claim')
-        return super(CrmClaim, self).copy(default)
+        results = self.env['crm.claim']
```
guewen: more simple than that: copy is not supposed to be called on several records, it is designed to copy only one record at a time, the code can be reduced to:

``` python
@api.multi
def copy(self, default=None):
    self.ensure_one()
    if default is None:
        default = {}
    default.setdefault('code', self.env['ir.sequence'].next_by_code('crm.claim'))
    return super(CrmClaim, record).copy(default)
```

crm_claim_code/models/crm_claim.py
guewen, line 32: my example above has one gotcha, it will increment the sequence even if code was already set, slightly better:

``` python
if 'code' not in default:
    default['code'] = self.env['ir.sequence'].next_by_code('crm.claim')
```

crm_claim_code/tests/test_crm_claim_code.py
```diff
@@ -34,12 +34,14 @@ def test_copy_claim_code_assign(self):
         self.assertNotEqual(crm_claim_copy.code, self.crm_claim.code)
         self.assertEqual(crm_claim_copy.code, code)
 
+    def test_copy_multi(self):
```
guewen: Is not supposed to be supported ;-)

cyrilgdn:
@guewen @pedrobaeza I changed copy method (to remove multi-copy support) and used environments context manager.

guewen:
Thanks! :+1:

crm_claim_code/__init__.py
```diff
@@ -15,11 +16,14 @@ def create_code_equal_to_id(cr):
...
-        cr.execute('UPDATE crm_claim '
-                   'SET code = \'%s\' '
-                   'WHERE id = %d;' %
-                   (sequence_obj.get(cr, SUPERUSER_ID, 'crm.claim'), claim_id))
+    with Environment.manage():
+        env = Environment(cr, SUPERUSER_ID, {})
+
+        sequence_obj = env['ir.sequence']
+
+        claim_ids = env['crm.claim'].search([], order="id")
```
pedrobaeza: Rename this to `claims`, as it is a recordset, not ids.

crm_claim_code/__init__.py
```diff
@@ -15,11 +16,14 @@ def create_code_equal_to_id(cr):
...
+        for claim_id in claim_ids:
+            cr.execute(
```
pedrobaeza: Why using here `cr.execute` if you have already set an environment? Make:

```
for claim in claim_ids:
     claim.code = sequence_obj.next_by_code('crm.claim')
```

cyrilgdn: @pedrobaeza Right sorry ! I changed it.

crm_claim_code/__init__.py
```diff
@@ -15,11 +16,9 @@ def create_code_equal_to_id(cr):
...
-        cr.execute('UPDATE crm_claim '
-                   'SET code = \'%s\' '
-                   'WHERE id = %d;' %
-                   (sequence_obj.get(cr, SUPERUSER_ID, 'crm.claim'), claim_id))
+    with Environment.manage():
+        env = Environment(cr, SUPERUSER_ID, {})
+
+        claims = env['crm.claim'].search([], order="id")
+        for claim in claims:
+            claim.code = env['ir.sequence'].next_by_code('crm.claim')
```
pedrobaeza: Why don't you store now the sequence handler in `sequence_obj` variable like before, so you avoid to search for it on each loop pass?

pedrobaeza:
:+1:

crm_claim_code/__init__.py
cyrilgdn: @pedrobaeza No real reason :-), except maybe reduction of code ! I reverted this change even if the cost of searching the model is really negligible against updating database row.

pedrobaeza: Yeah, I know it's less, but if we can win some performance, why not do it?

leemannd:
:+1:

PR #95, opened 2016-05-19
[8.0][crm_lead_lost_reason] Add button Mark Lost in view CRM - Leads Form

ghost:
The "Mark Lost" button was not in the form view of the leads, only in the view of opportunities.
It's necessary to establish as lost without convert to opportunity.


rafaelbn:
Hi @rr-clearcorp , I don't think like you because a lead is to far away to be a customer so it's enough to have a Stage "Dead" (or whatever you want to name it). But you don't need a reason. If you need a reason it means that it must be an opportunity and you continue with the possibility of losing it. If you don't want to have 2 steps, then start from opportunity and not from Lead.

Any other opinion?

Thanks

pedrobaeza:
I think the same as @rafaelbn, so I close this one.

PR #96, opened 2016-05-30
[IMP][8.0][mass_mailing_partner] Sync opt_out.

yajo:
If you create or update a contact that is linked to a partner and set a value for opt_out, then that value will be set on the partner too.

@Tecnativa


pedrobaeza:
No, this sync shouldn't be done. You can put an opt_out in a mass mailing contact because you don't want to be in a certain list, but the full partner can still receive mails from other lists.

:-1:

PR #99, opened 2016-06-09
[ADD] crm_deduplicate_by_fuzzy_name

pedrobaeza:
# 
# Deduplicate Contacts by similar name

This module extends the criteria to match duplicated contacts looking for
similar names according a maximum of different characters.
# Installation

You need to have the _fuzzystrmatch_ Postgres extension installed. You can
add it from a psql command line typing:

   CREATE EXTENSION fuzzystrmatch;
# Usage

- Go to _Sales > Tools > Deduplicate Contacts_.
- Mark "Similar name" in the section "Search duplicates based on duplicated
  data in", and select the maximum number of different characters that you
  are going to allow.
- This criteria will be used for deduplicating.

@Tecnativa


pedrobaeza:
cc @Tecnativa

pedrobaeza:
Uhm, here I need to install the extension in the DB...

crm_deduplicate_by_fuzzy_name/README.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Installation
+============
+
+You need to have the *fuzzystrmatch* Postgres extension installed. You can
+add it from a psql command line typing:
+
+.. code-block:: psql
+
+   CREATE EXTENSION fuzzystrmatch;
```
yajo, line 20: Is it possible to make the module execute this? And to add this to binary dependencies? I think some extra package is needed that may not be found in all Postgres installations.

crm_deduplicate_by_fuzzy_name/security/crm_deduplicate_acl_security.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- © 2016 Antiun Ingeniería S.L. - Jairo Llopis
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+
+<openerp>
+<data>
+
+<!-- New user group category -->
+<record id="category" model="ir.module.category">
+    <field name="name">Deduplicate Contacts</field>
```
yajo: You should depend on `crm_deduplicate_acl` instead.

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
```diff
@@ -0,0 +1,129 @@
...
+            """)
+        if not cr.fetchall():
+            try:
+                # Try to install the extension
+                cr.execute("CREATE EXTENSION fuzzystrmatch")
+            except:
+                raise UserError(
+                    _("PostgreSQL extension 'fuzzystrmatch' is not present on "
+                      "your DB and can't be auto-installed. Please install "
+                      "it."))
```
yajo, line 37: Does this error get to the user, or is it only in the logs?

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
```diff
@@ -0,0 +1,129 @@
...
+        return super(BasePartnerMergeAutomaticWizard,
+                     self)._register_hook(cr)
+
+    def _compute_selected_groupby(self, this):
+        if this.fuzzy_name_group_by:
+            try:
+                return super(BasePartnerMergeAutomaticWizard,
+                             self)._compute_selected_groupby(this)
+            except osv.except_osv:
```
yajo, line 46: Use `exceptions.except_osv`; `osv` is a deprecated module.

crm_deduplicate_by_fuzzy_name/security/crm_deduplicate_acl_security.xml
pedrobaeza: Sorry, this is a copy/paste error from that module.

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
pedrobaeza, line 37: It gets the user. The problem is that it tries to install again and again until solved. I haven't found any better solution.

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
pedrobaeza, line 46: I know, but it's the exception that is raised upstream

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
```diff
@@ -0,0 +1,129 @@
...
+                self.env.cr.execute(
+                    "SELECT id FROM res_partner "
+                    "WHERE active = True AND (%s)" % where_query)
+                partners = partner_obj.browse(
+                    [x[0] for x in self.env.cr.fetchall()])
+            else:
+                partners = partner_obj.search([])
+            while partners:
+                partner = partners[0]
+                partners -= partner
```
yajo, line 77: Although this probably works, `for partner in partners` is more pythonic.

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
```diff
@@ -0,0 +1,129 @@
...
+                    "WHERE levenshtein("
+                    "  %s, "
+                    "  substring(name for 100))"
+                    "  <= %s "
+                    "AND id != %s")
+                for field in groups:
+                    query += " AND %s = '%s'" % (
+                        field, getattr(partner, field))
+                self.env.cr.execute(
+                    query, (partner.name[:100],
```
yajo, line 93: Why only the 100 first characters of the name?

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
pedrobaeza, line 77: No, because this is a dynamic list (see later `partners -= partner_obj.browse(match_partner_ids)`)

yajo, line 77: OK forget it, I did not see https://github.com/OCA/crm/pull/99/files#diff-fb3d8874d5ff4e6cd2db4396dbd813bbR94

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
pedrobaeza, line 93: To ensure levenshtein don't give an error (it has a limit).

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
```diff
@@ -0,0 +1,129 @@
...
+    _inherit = "base.partner.merge.automatic.wizard"
+
+    # The field is named this way to not be automatically added as a "group by"
+    # option, which is performed in the code through the name of the field
+    fuzzy_name_group_by = fields.Boolean(
+        string='Similar name',
+        help="WARNING: This option is significantly slower than the others, "
+             "because the match cannot be performed in group.")
+    max_fuzzy_name_difference = fields.Integer(
+        string="Max. characters difference", default=2)
```
yajo, line 20: Have you considered making this a percent value? Maybe I'm saying an stupidity, but if a name is 5 characters long and 2 differ, seems equivalent to 20 characters in a 100-char-long name.

Setting it to 2% (rounded up) would make it more consistent IMHO.

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
yajo, line 37: Probably a `pre_init_hook` would work.

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
yajo, line 46: [It's the same.](https://github.com/odoo/odoo/blob/8.0/openerp/osv/osv.py#L27)

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
yajo, line 93: Wouldn't it be better to just exclude partners with name longer than 100 from the process? This could lead to possible mistakes.

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
pedrobaeza, line 37: Uhm, good idea. I'll try.

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
pedrobaeza, line 20: I considered it, but as at the end, I have to "translate" it to characters difference, I see it less accurate. What is a 2% difference in a 5 letter word?

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
pedrobaeza, line 46: No, Python doesn't consider the same and it's not caught

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
pedrobaeza, line 93: I don't think so. If the first 100 characters are similar, they will probably be the same. We are talking about the name, that will hardly overpass this length.

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
yajo, line 20: rounded up, 1 char.

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
yajo, line 46:
```
>>> ex = Exception
>>> try:
...     raise ex
... except Exception:
...     print "hi"
... 
hi
```

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
pedrobaeza, line 20: Yeah, but from user perspective, it's even more confuse

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
pedrobaeza, line 46: But I don't want to catch all exceptions. Only that one.

yajo, line 46: I mean that `osv.except_osv` is exactly the same memory object than `exceptions.except_orm`. Exception will be caught.

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
yajo, line 20: Well no problem, it was just an idea :wink:

crm_deduplicate_by_fuzzy_name/wizards/partner_merge.py
pedrobaeza, line 46: OK, I'll try

rafaelbn:
Is travis error related? Are you going to fix it?

crm_deduplicate_by_fuzzy_name/__openerp__.py
```diff
@@ -0,0 +1,25 @@
...
+    "version": "8.0.1.0.0",
+    "category": "Tools",
+    "website": "https://www.tecnativa.com",
+    "author": "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "external_dependencies": {
+        'python': [
+            'unidecode',
```
yajo: This kills travis and seems you don't actually use it in your code.

pedrobaeza:
@rafaelbn, it's difficult, because it requires DB admin for installing the PostgreSQL extension I use. I need to check if I can hack this in any way in Travis.

crm_deduplicate_by_fuzzy_name/__openerp__.py
pedrobaeza: :open_mouth: I thought I remove it!

rafaelbn:
OK, I'll wait.

rafaelbn:
Rebuilding for testing, ping @carlosdauden

rafaelbn:
Travis fails. Do you know why?

I've tested this module in 2 scenarios:

- Runbot and 3 res.partners with similar names --> OK, works
- Real customer database with 30K customers, never give me a result to merge. I've wait in 5 different times for more than 5 hours and no results, so in this case is unusable.

[ ] Still pending a test with a database with for example 10K customer to check if it works or not.

cc @carlosdauden

yajo:
I think @pedrobaeza answered you just above your question...

pedrobaeza:
Closing as discarded due to performance reasons. Another alternative method was used.

PR #101, opened 2016-06-27
[ADD] Add crm_sale_open_editable

mistotebe:
When a quotation is created from an opportunity, it is far from done yet (there are no lines for example), so it's quite useful for the user if it opens in edit mode immediately.

This module does just that.


rafaelbn:
Hi @mistotebe , thanks for your contribution but what this module makes is just saving 1 click. I don't think we must have in OCA modules for everything. It my opinion. Any other opinions?

mistotebe:
@rafaelbn I don't dispute that, just that we got this request from a few sales people and they found it useful.

pedrobaeza:
I agree this one is very short for a module. We usually include this kind of things in the customer custom module, or better, try to include this little fix on Odoo core/OCB. You can propose a fix to them.

I close this meanwhile. If there's interest on it, please say it and I'll reopen again.

PR #102, opened 2016-06-29
Fix creation of existing column

guewen:
When a database already have this field, the module cannot be installed.
Now it only creates the code column if it doesn't exist yet.

### DIFF
--- a/crm_claim_code/__init__.py
+++ b/crm_claim_code/__init__.py
@@ -9,10 +9,13 @@ from openerp import SUPERUSER_ID
 
 
 def create_code_equal_to_id(cr):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

rafaelbn:
:+1:

cyrilgdn:
:+1:

guewen:
@pedrobaeza @rafaelbn @cyrilgdn 
Thanks for your reviews. I added a flag to know whether the post operation `assign_old_sequences` should be executed or not, otherwise it overrides the existing values in the `code` field

PR #105, opened 2016-07-13
[FIX] only users beloning to group 'Deduplicate Contacts / Manually' should be able to merge partners from the partner list views (action 'Merge automatically')

JordiBForgeFlow:
Now any partner can merge partners from any partner tree view (such as 'Sales / Customers').

With this fix, a user will only be able to see the option 'Merge automatically' in the partners tree view if he belongs to group "Deduplicate Contacts / Manually"

### DIFF
--- a/crm_deduplicate_acl/wizards/partner_merge_view.xml
+++ b/crm_deduplicate_acl/wizards/partner_merge_view.xml
@@ -3,29 +3,30 @@
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
 
 <openerp>
-<data>
+    <data>
 
-<record id="base_partner_merge_automatic_wizard_form" model="ir.ui.view">
-    <field name="name">Restrict automatic merge access</field>
-    <field name="model">base.partner.merge.automatic.wizard</field>
-    <field
-        name="inherit_id"
-        ref="crm.base_partner_merge_automatic_wizard_form"/>
-    <field name="arch" type="xml">
-        <xpath expr="//button[@name='automatic_process_cb']"
-               position="attributes">
-            <attribute name="groups">
-                crm_deduplicate_acl.group_automatically
-            </attribute>
-        </xpath>
-        <xpath expr="//button[@name='update_all_process_cb']"
-               position="attributes">
-            <attribute name="groups">
-                crm_deduplicate_acl.group_automatically
-            </attribute>
-        </xpath>
-    </field>
-</record>
+        <record id="base_partner_merge_automatic_wizard_form" model="ir.ui.view">
+            <field name="name">Restrict automatic merge access</field>
+            <field name="model">base.partner.merge.automatic.wizard</field>
+            <field
+                name="inherit_id"
+                ref="crm.base_partner_merge_automatic_wizard_form"/>
+            <field name="arch" type="xml">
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
@rafaelbn @pedrobaeza @Yajo you might be interested in this.

JordiBForgeFlow:
The PR also fixes an issue by which the buttons 'automatic_process_cb' and 'update_all_process_cb' were not visible.

crm_deduplicate_acl/wizards/partner_merge_view.xml
```diff
@@ -3,29 +3,30 @@
...
-        </xpath>
-        <xpath expr="//button[@name='update_all_process_cb']"
-               position="attributes">
-            <attribute name="groups">
-                crm_deduplicate_acl.group_automatically
-            </attribute>
-        </xpath>
-    </field>
-</record>
+        <record id="base_partner_merge_automatic_wizard_form" model="ir.ui.view">
```
pedrobaeza, line 8: Don't reformat XML, as this makes more difficult to identify the changes

crm_deduplicate_acl/wizards/partner_merge_view.xml
```diff
@@ -3,29 +3,30 @@
...
+                </xpath>
+                <xpath expr="//button[@name='update_all_process_cb']"
+                       position="attributes">
+                    <attribute name="groups">crm_deduplicate_acl.group_automatically</attribute>
+                </xpath>
+            </field>
+        </record>
 
-</data>
+        <act_window id="crm.action_partner_merge"
```
pedrobaeza: Use `<record id="crm.action_partner_merge"...` to change only affected field.

crm_deduplicate_acl/wizards/partner_merge_view.xml
```diff
@@ -3,29 +3,27 @@
...
+                </xpath>
+                <xpath expr="//button[@name='update_all_process_cb']"
+                       position="attributes">
+                    <attribute name="groups">crm_deduplicate_acl.group_automatically</attribute>
+                </xpath>
+            </field>
+        </record>
 
-</data>
+        <act_window id="crm.action_partner_merge" groups="group_manually"/>
```
pedrobaeza: It must be `<record>`, not `<act_window>`. This doesn't work.

crm_deduplicate_acl/wizards/partner_merge_view.xml
pedrobaeza, line 8: Please don't change this.

JordiBForgeFlow, line 8: At least in my local computer if I do not change that, the application fails to recognize the groups. See how I added the permissions to demo user:
image

But then when I open the wizard for deduplication, the automatic merge buttons are not visible:
image

When I re-format the XML, they do show up correctly.

pedrobaeza, line 8: OK for not adding line breaks on the attribute itself, but please don't indent all of them.

pedrobaeza:
:+1: when Travis goes green

yajo:
:+1: Good fix. IMHO you should squash (it's a little diff for 5 commits), but not a blocker.

rafaelbn:
Tested :+1:

PR #109, opened 2016-08-12
[ADD] crm_deduplicate_filter

pedrobaeza:
# 
# Exclude some records from the deduplication

This module extends the possibilities of the contact deduplication allowing
to filter the applicable set according to several criteria.

For now, only the filter for restricting the deduplication to only companies or
only contacts is implemented.
# Usage

   "Parent company set (Contacts)" in the section 'Exclude contacts having'.
   kind of records.


pedrobaeza:
cc @Tecnativa

crm_deduplicate_filter/README.rst
```diff
@@ -0,0 +1,65 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===========================================
+Exclude some records from the deduplication
+===========================================
+
+This module extends the possibilities of the contact deduplication allowing
+to filter the applicable set according several criteria.
```
elicoidal: according to

elicoidal:
:+1:

crm_deduplicate_filter/README.rst
```diff
@@ -0,0 +1,65 @@
...
+
+===========================================
+Exclude some records from the deduplication
+===========================================
+
+This module extends the possibilities of the contact deduplication allowing
+to filter the applicable set according to several criteria.
+
+For now, only the filter for restricting the deduplication to only companies or
+only contacts is implemented.
```
yajo, line 13: This should go to known issues instead

yajo:
:+1:

rafaelbn:
@pedrobaeza if you review comments we can merge this one

crm_deduplicate_filter/README.rst
pedrobaeza: That's not a known issue, but only the scope. The known issue / roadmap should be "Implement more filters for deduplication", but I decided to let it this way.

PR #110, opened 2016-09-06
[FIX][mass_mailing_partner] Exclude opt_out.

yajo:
Now opted-out records will not be counted in the "Mailing lists" smart button
in the partner form.

@Tecnativa

### DIFF
--- a/mass_mailing_partner/README.rst
+++ b/mass_mailing_partner/README.rst
@@ -82,6 +82,7 @@ Contributors
 * Rafael Blasco <rafabn@antiun.com>
 * Antonio Espinosa <antonioea@antiun.com>
 * Javier Iniesta <javieria@antiun.com>
+* Jairo Llopis <jairo.llopis@tecnativa.com>
 
 Maintainer
 ----------
--- a/mass_mailing_partner/README.rst
+++ b/mass_mailing_partner/README.rst
@@ -79,7 +79,7 @@ Contributors
 ------------
 
 * Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
-* Rafael Blasco <rafabn@antiun.com>
+* Rafael Blasco <rafael.blasco@tecnativa.com>
 * Antonio Espinosa <antonioea@antiun.com>
 * Javier Iniesta <javieria@antiun.com>
 * Jairo Llopis <jairo.llopis@tecnativa.com>
--- a/mass_mailing_partner/__openerp__.py
+++ b/mass_mailing_partner/__openerp__.py
@@ -7,7 +7,7 @@
 
 {
     "name": "Link partners with mass-mailing",
-    "version": "8.0.3.0.0",
+    "version": "8.0.2.1.0",
     "author": "Tecnativa, "
               "Antiun Ingeniería S.L., "
               "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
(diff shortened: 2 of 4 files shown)


mass_mailing_partner/__openerp__.py
```diff
@@ -7,7 +7,7 @@
 
 {
     "name": "Link partners with mass-mailing",
-    "version": "8.0.2.0.0",
+    "version": "8.0.3.0.0",
```
pedrobaeza: Version should be 8.0.2.1.0

mass_mailing_partner/models/res_partner.py
```diff
@@ -52,5 +54,5 @@ def write(self, vals):
                 mm_vals['name'] = vals['name']
             if vals.get('email'):
                 mm_vals['name'] = vals['email']
-            self.mapped('mass_mailing_contacts').write(mm_vals)
+            self.mapped('mass_mailing_contact_ids').write(mm_vals)
```
pedrobaeza: This change should be done in all contacts, with or without opt_out

mass_mailing_partner/__openerp__.py
yajo: I changed a field name, which implies database change and module update, that's why I raised to 3.0.0.

pedrobaeza: No, it's not a big change and even doesn't require migration script, so put 2.1.0

pedrobaeza:
:+1:

mass_mailing_partner/models/res_partner.py
```diff
@@ -28,15 +30,15 @@ class ResPartner(models.Model):
...
     def _check_email_mass_mailing_contacts(self):
-        if self.mass_mailing_contacts and not self.email:
+        if self.mass_mailing_contact_ids and not self.email:
             raise ValidationError(
                 _("This partner '%s' is subscribed to one or more "
                   "mailing lists. Email must be assigned." % self.name))
 
     @api.one
-    @api.depends('mass_mailing_contacts')
+    @api.depends('mass_mailing_contact_ids')
```
antespi: You have to add `mass_mailing_contact_ids.opt_out` in order to recompute when one contact change from `opt_out = False` to `opt_out = True`

Use case to reproduce:
1. Go to Agrolait partner form
2. Click on Mailing lists smart button (count = 0)
3. Create a contact, Agrolait will be the partner related with it automatically
4. Return to partner form. Mailing lists smart butoon now shows count = 1
5. Click again on Mailing lists smart button
6. Change opt_out column from False to True
7. Return to partner fomr. Mailing lists smart button shows count = 1, but it should show count = 0

To fix this, change api.depends:

```
@api.depends('mass_mailing_contact_ids', 'mass_mailing_contact_ids.opt_out')
```

gurneyalex: good catch!

mass_mailing_partner/README.rst
```diff
@@ -82,6 +82,7 @@ Contributors
 * Rafael Blasco <rafabn@antiun.com>
```
rafaelbn: rafael.blasco@tecnativa.com

rafaelbn:
@Yajo please check comments

antespi:
:+1:

PR #116, opened 2016-10-05
lettermgmt support 9

surenjie:
odoo9 non existent ir.sequence.type model


pedrobaeza:
There's a refactoring in course for v8, so better to open this after the merge of this refactor. I close meanwhile.

PR #35, opened 2015-07-06
[ADD] Added calendar_resources module.

bjoliveau:
This module has been developed by Jordi @foutoucour 

I've just discovered it hasn't been merged on OCA and think it's a nice feature, used by the customer who financed it.

This module add new features to facilitate your resources management with your meetings :
- improvement in resource.resource native object (image, note and boolean choice for using it in your calendars)
- add a link between your calendars and your resources,
- switch easily your calendar's views with attendees and resources,
- add a new menu entry to display event by resources in several views (calendar, list and form)

Note that the customer don't expect for this step :
- any constraint between resources reservations in calendar,
- any interface with calendar_google for the 'resource' object.

PR #118, opened 2016-10-17
Add calendar_resource module, for meeting rooms

dreispt:
Continuing #35


dreispt:
cc @bjoliveau

dreispt:
This is not CRM specific.
I'm considering renaming and moving it to OCA/social.
How about `calendar_resource` for name?

lasley (changes requested):
Code looks good. It would be nice if the headers were all reduced & a ReadMe added though.

I agree with the rename, that's exactly what I was googling to try to find a module such as this.

I'm not so sure if OCA/social is a good repo though. Neither is `event`. We don't have a `calendar` or `resource` repo either. So yeah I don't have a better suggestion there 😦

pedrobaeza:
If you're going to create a bunch of resource related modules, it worths to create OCA/resource.

lasley:
I'll likely have one or two on top of this as part of OCA/product-attribute#206 & possibly a few more in the coming months. My goal for v10 iteration of vertical-medical is to remove and abstract as much of the logic as possible. There's a decent amount related to resources that I haven't looked at yet too such as hospital beds, operating centers, etc.

calendar_resource/__openerp__.py
```diff
@@ -0,0 +1,65 @@
...
+    'name': 'Calendar Resources',
+    'version': '8.0.1.0.1',
+    'author': 'Savoir-faire Linux',
+    'maintainer': 'Savoir-faire Linux',
+    'website': 'http://www.savoirfairelinux.com',
+    'license': 'AGPL-3',
+    'category': 'CRM',
+    'summary': 'New features to facilitate resources management with meetings',
+    'description': """
+CRM Meeting Resources
```
lasley, line 33: Maybe we can abstract this description a bit? I plan on using this for everything from Physician scheduling to Operating Rooms

dreispt: Appointments? Calendar Entries?

lasley: I think `Appointments` would work well. Calendar Entries may cause some confusion with Calendar Events. Thanks @dreispt

lasley:
@dreispt - I'm going to need this soon for medical in v10. Are you planning on making any more changes soon, or would you like me to just handle the rest of this PR for version 10 in whatever repo we're actually going to put it in?

pedrobaeza:
Superseeded by #135

lasley:
@dreispt - We have continued this PR in v10 in #135. Unfortunately anything below 10 is out of our scope, so we won't be backporting. That said though, I recommend that we still close this PR in favor of the complete one. If someone wants it in an older version, it should be a simple backport.

dreispt:
Go ahead with #135

PR #121, opened 2016-10-31
[IMP] rename table instead of copying

hbrunn:
In https://github.com/OCA/crm/pull/87#issuecomment-247669240, @ivantodorovich talked about this issue, but we didn't followup on it. We really need to rename the table instead of selecting stuff, because otherwise things go very wrong when other modules add columns. Also, we can't assume that column order in both tables is the same.

Further, we don't need to drop the indices as they'll be dropped automatically when the corresponding column is dropped.

### DIFF
--- a/lettermgmt/migrations/8.0.2.0/post-migration.py
+++ b/lettermgmt/migrations/8.0.2.0/post-migration.py
@@ -20,22 +20,7 @@ def migrate(cr, version):
     # Rename column 'type' to 'type_id'
     cr.execute("""UPDATE res_letter SET type_id = type""")
 
-    # Rename model 'letter.class' to 'letter.category'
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

ivantodorovich:
Have you tested if it migrates ok with data?
Because I remember trying that approach, and it throwed some errors on migration.

Although, may have been corrupted data on my test database..

hbrunn:
@ivantodorovich yes, this is tested. I assume you wrote the migration script after your development, then of course a table with this name exists already, and it will break

ivantodorovich:
It's possible. I remember that at some point, I got very lost and didn't even knew what I was doing.

Anyway, 👍 +1

yajo (approved):

PR #122, opened 2016-11-07
9.0 [MIG] crm_deduplicate_acl migration

cubells:


pedrobaeza:
Check Travis error:

```
crm_deduplicate_acl/wizards/partner_merge.py:22:29: E131 continuation line unaligned for hanging indent
```

crm_deduplicate_acl/wizards/partner_merge.py
```diff
@@ -0,0 +1,25 @@
...
+        """Allow non-admins to merge partners with different emails."""
+        # Know if user has unrestricted access
+        group_unrestricted = self.env["ir.model.data"].xmlid_to_object(
+            "crm_deduplicate_acl.group_unrestricted")
+
+        # Run as admin if so
+        return super(BasePartnerMergeAutomaticWizard, self)._merge(
+            self.env.cr,
+            SUPERUSER_ID if group_unrestricted.id in self.env.user.groups_id
+                .ids else self.env.uid,
```
yajo: I think you should use `self.env.user.has_group` instead.

yajo (changes requested):

pedrobaeza (approved):

yajo (approved):
Code review.

rafaelbn (approved):
Thanks

rafaelbn:
@cubells one question, where is the functionality of "deduplicate" in v9? in which menu? thanks

rafaelbn (dismissed):
@cubells one question, where is the functionality of "deduplicate" in v9? in which menu? thanks

cubells:
v8.0 menu doesn't exist in version 9.0.

Module now only inherits wizard to merge partner.

rafaelbn:
I think we should recover this menu item. This is a functionality that without this menu item lost a lot.

cubells:
I can create a new menu if you want, as the parent menu of version 8.0 has disappear in v.9.0.

@pedrobaeza @Yajo what do you think?

cubells:
Sorry Rafa: menu disappear in crm module. I think It's better do a PR in crm module:

https://github.com/OCA/OCB/blob/8.0/addons/crm/base_partner_merge_view.xml#L16

rafaelbn:
I agree @cubells with a PR to CRM module. Let see what @pedrobaeza thinks about this

pedrobaeza:
It was removed here https://github.com/odoo/odoo/commit/30a7ef804a1457fb0274994993c86a01dbb114d4#diff-be356a586330623cb55dac5cdf755b17L15 without apparent reason. I have asked if Odoo would accept a PR amending this in a commit that fixes some things of that commit: https://github.com/odoo/odoo/commit/876634b1dabb64414df1558d4ef0e99db8185f73.

If not, we should add here that menu, and depend on this one for other modules that improves deduplication wizard features (like _crm_deduplicate_by_website_ or  _crm_deduplicate_filter_).

pedrobaeza:
Please include the menu as there's no answer.

Issue #125, opened 2016-12-28
[9.0] crm_deduplicate_acl error


edser-pcg:
When trying to use this module in v9 it returns the following error after you click on Combine Contacts on either option (manual, auto), which doesn't returns the error when it isn't installed. Privileges to user where all set.

Wanted to try this module to see if it would eliminate the limit of up to 3 contacts to merge at the same time.

```
Traceback (most recent call last):
  File "/odoo/odoo-server/src/odoo/openerp/http.py", line 648, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/odoo/odoo-server/src/odoo/openerp/http.py", line 685, in dispatch
    result = self._call_function(**self.params)
  File "/odoo/odoo-server/src/odoo/openerp/http.py", line 321, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/odoo/odoo-server/src/odoo/openerp/service/model.py", line 118, in wrapper
    return f(dbname, *args, **kwargs)
  File "/odoo/odoo-server/src/odoo/openerp/http.py", line 314, in checked_call
    result = self.endpoint(*a, **kw)
  File "/odoo/odoo-server/src/odoo/openerp/http.py", line 964, in __call__
    return self.method(*args, **kw)
  File "/odoo/odoo-server/src/odoo/openerp/http.py", line 514, in response_wrap
    response = f(*args, **kw)
  File "/odoo/odoo-server/src/odoo/addons/web/controllers/main.py", line 892, in call_button
    action = self._call_kw(model, method, args, {})
  File "/odoo/odoo-server/src/odoo/addons/web/controllers/main.py", line 880, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/odoo/odoo-server/src/odoo/openerp/api.py", line 250, in wrapper
    return old_api(self, *args, **kwargs)
  File "/odoo/odoo-server/src/odoo/addons/crm/base_partner_merge.py", line 744, in merge_cb
    self._merge(cr, uid, partner_ids, this.dst_partner_id, context=context)
  File "/odoo/odoo-server/src/odoo/openerp/api.py", line 250, in wrapper
    return old_api(self, *args, **kwargs)
  File "/odoo/odoo-server/src/odoo/openerp/api.py", line 354, in old_api
    result = method(recs, *args, **kwargs)
[...]

PR #128, opened 2017-01-11
[Fix] Keyword arguments problem on merge partner

cubells:
- Fix issue #125 
cc @Tecnativa


pedrobaeza:
Why codecov complains that this part is not tested?

yajo:
I guess because it was not tested before.

yajo (approved):

pedrobaeza (approved):
Merging for fixing the problem. The ideal would be to have this tested.

PR #130, opened 2017-01-20
[10.0] [ADD] Crm lead followers

LeartS:
# Share leads with followers

It's sometimes useful to share leads with some users that would normally not have
access to them, e.g. developers for a service company so that can see and offer
their input in possibile future projects they'll have to work on.
Unfortunately the default access level on the CRM and Sale applications do not
allow for this: either a user can see all leads, or he can see only his own
leads, there is no way to "share" leads with non-sale users.

This module solves this by extending the permissions of the CRM module so that users
can be put in a group where they have (read or write) access only on leads that
were explicitly shared with them (by adding them as followers).

### DIFF
--- a/crm_lead_followers/README.rst
+++ b/crm_lead_followers/README.rst
@@ -1,0 +1,84 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==========================
+Share leads with followers
+==========================
+
+It's sometimes useful to share leads with some users that would normally not have
+access to them, e.g. developers for a service company so that can see and offer
+their input in possibile future projects they'll have to work on.
+Unfortunately the default access level on the CRM and Sale applications do not
+allow for this: either a user can see all leads, or he can see only his own
+leads, there is no way to "share" leads with non-sale users.
+
+This module solves this by extending the permissions of the CRM module so that users
+can be put in a group where they have (read or write) access only on leads that
+were explicitly shared with them (by adding them as followers).
+
+Configuration
+=============
+
+This module defines two additional access levels for the sale/crm application:
+
+* See Shared Leads: users in this groups can see leads where they have been added as followers
+* Edit Shared Leads: users in this groups can see and edit leads where they have been added as followers
+
+Neither of these groups give the user the permission to create or delete leads, nor they impact
+permissions on sale orders and quotations.
+Users of the group 'Sale / Own Documents Only' are automatically added to the 'Edit Shared Leads' group,
+that means that they will be automatically able to edit leads they are following.
+
+
+Usage
+=====
+
+Add the users to the followers of the leads you want them to be able to see/edit
+(according to the group you put them in).
+
+Known issues / Roadmap
+======================
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/crm/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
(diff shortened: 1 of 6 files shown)


crm_lead_followers/security/crm_lead_followers_security.xml
```diff
@@ -0,0 +1,34 @@
...
+  </record>
+
+  <record id="sales_team.group_sale_salesman" model="res.groups">
+    <field name="implied_ids" eval="[(4, ref('crm_lead_followers.group_edit_shared_leads'))]"/>
+  </record>
+
+  <record id="crm_lead_rule_see_shared_leads" model="ir.rule">
+    <field name="name">See followed Leads</field>
+    <field name="model_id" ref="crm.model_crm_lead"/>
+    <field name="domain_force">[('message_partner_ids','in',[user.partner_id.id])]</field>
```
Fenkiou, line 23: Missing spaces after commas

crm_lead_followers/security/crm_lead_followers_security.xml
```diff
@@ -0,0 +1,34 @@
...
+    <field name="groups" eval="[(4, ref('group_see_shared_leads'))]"/>
+  </record>
+
+  <record id="crm_lead_rule_edit_shared_leads" model="ir.rule">
+    <field name="name">Edit followed Leads</field>
+    <field name="model_id" ref="crm.model_crm_lead"/>
+    <field name="domain_force">[('message_partner_ids','in',[user.partner_id.id])]</field>
```
Fenkiou, line 30: Missing spaces after commas

Fenkiou (commented):
Would it be possible to add some unit tests ?

- One to check that non follower user can't see a lead.
- One to check that user added as follower can't see a lead.
- One to check that a user added as follower and in the group `See Shared Leads` can see a lead but can't edit it.
- One to check that a user added as follower and in the group `Edit Shared Leads` can see a lead and edit the lead.

I do not know CRM module, so maybe some tests are irrelevant.

crm_lead_followers/README.rst
```diff
@@ -0,0 +1,84 @@
...
+Users of the group 'Sale / Own Documents Only' are automatically added to the 'Edit Shared Leads' group,
+that means that they will be automatically able to edit leads they are following.
+
+
+Usage
+=====
+
+Add the users to the followers of the leads you want them to be able to see/edit
+(according to the group you put them in).
+
```
misern2, line 39: Runbot section is missing.

misern2 (commented):
Fast code Review

misern2:
@LeartS  Please commit this littles changes to merge ASAP.

Thanks.

rafaelbn:
Please @LeartS could you attend comments? Thanks! 😄

crm_lead_followers/security/crm_lead_followers_security.xml
yajo: You can use `.ids` instead of the array

crm_lead_followers/views/crm.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+  <record id="crm_lead_view_search" model="ir.ui.view">
+    <field name="name">lead search view: add shared leads to my opportunities</field>
+    <field name="model">crm.lead</field>
+    <field name="inherit_id" ref="crm.view_crm_case_opportunities_filter"/>
+    <field name="arch" type="xml">
+      <filter name="assigned_to_me" position="attributes">
+        <attribute name="domain">['|', ('user_id', '=', uid), ('message_is_follower', '=', True)]</attribute>
```
yajo, line 10: I think that the "assigned to me" filter should not include every lead you follow. That would be misleading.
It's better to either add a new filter or just add nothing, since anyways permissions will filter out those that you don't follow (and thus have no access to).

yajo (approved):
Code looks OK, couldn't test, waiting for runbot. Just a few comments.

However I have another suggestion: do you think it would make sense to grant access to leads where there's a channel assigned to which you belong? (At the end, you're a follower too).

PR #140, opened 2017-05-02
[MIG][9.0] crm_action module

cubells:
- Update README
- Update views
- Add tests

cc @Tecnativa


crm_action/__openerp__.py
```diff
@@ -1,15 +1,20 @@
...
+# Copyright 2017 Tecnativa - Vicent Cubells
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
     'name': 'CRM Action',
-    'version': '8.0.1.2.0',
-    'author': 'Savoir-faire Linux,Odoo Community Association (OCA)',
+    'version': '9.0.1.0.0',
+    'author': 'Savoir-faire Linux, '
+              'Tecnativa, '
```
dreispt: I know that @pedrobaeza disagrees, but porting does not grant authorship/copyright.

crm_action/data/email_reminder.xml
```diff
@@ -1,36 +1,34 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!--
-  © 2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+  Cpyright 2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+  Cpyright 2017 Tecnativa - Vicent Cubells
```
dreispt: s/Cpyright /Copyright

crm_action/models/crm_lead.py
```diff
@@ -1,16 +1,44 @@
...
 
-from openerp import models, fields, api
+from openerp import api, fields, models
 
 
 class CrmLead(models.Model):
     _inherit = 'crm.lead'
 
-    def count_actions(self):
+    def compute_actions(self):
```
dreispt: I think `compute_count_actions` would be more explicit, but not an issue.

dreispt (commented):
Overall looks good, only some nitckpicks to look at.

crm_action/__openerp__.py
pedrobaeza: Again this is not a plain migration, but an enhancement including unit tests.

dreispt: I did some more thinking and won't complain on this anymore: for practical purposes, adding to the copyright has no real impact, and can be a way for contribution recognition.

pedrobaeza: I actually agree with you that some contributions don't deserve a place here. A migration always deserve it? Well, it depends. If you only follow the migration guide, and throw it here, maybe not. It even requires the check through the UI or unit tests... so it's not easy to set a general rule.

luismontalba (approved):
Functionality tested on runbot

dreispt (approved):

hbrunn (approved):

PR #142, opened 2017-05-11
[IMP] Add smartbutton to leads and opportunities

cubells:
Usability increase if smart button is added to leads and pipelines.

cc @Tecnativa

### DIFF
--- a/crm_phonecall/__openerp__.py
+++ b/crm_phonecall/__openerp__.py
@@ -24,6 +24,7 @@
         'wizard/crm_phonecall_to_phonecall_view.xml',
         'views/crm_phonecall_view.xml',
         'views/res_partner_view.xml',
+        'views/crm_lead_view.xml',
         'report/crm_phonecall_report_view.xml',
     ],
     'installable': True,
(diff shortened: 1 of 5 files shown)


rafaelbn (approved):
Tested functionally in runbot , thanks!

chienandalu:
Wouldn't it be better if the smart button had this behaviour?:

- 0-1 calls: goes to call form view (so you either schedule or register or, at least, save one click :D)
- >1 calls: goes to calls tree view as it does now.

And maybe not the focus of this PR: 

- Wouldn't it be better if Scheduled Calls menu would be available to all users by default? It's like a hidden feature until you realise that you have to enable permissions in the user's prefs.

- Form view available in both Logged calls and Scheduled calls view would improve the flow (i.e. if you hold a call by mistake it's hard to recover it to another one. This form view is only accesible form the calendar view and then clicking in the call in Logged calls view.

luismontalba (approved):
Functionality tested

chienandalu (approved):
Sorry, my last comment didn't go here. This is aproved

PR #147, opened 2017-05-15
crm_action module migration to v10

qtheuret:
Migration steps for v10:
- Add 'sales_team' module as dependencies
- Rewrite some xml files to get sales_team references instead of base ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ Available addons
 ----------------
 addon | version | summary
 --- | --- | ---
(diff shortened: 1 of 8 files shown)


pedrobaeza:
This has been recently migrated to 9.0, so you should follow migration guide for bringing 9.0 commits (and your migration will be lighter): https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0

PR #149, opened 2017-05-24
[10.0][ADD] calendar_resource: Add further resource validations.

woodbrettm:
### This PR adds the following to calendar_resource:
- [x] Restrict resources to certain calendar event types
- [x] Validate resource leaves on events
- [x] Prevent resources from being added to appointments outside of their availability times

### Looking for Feedback on:

This PR is partially done, however after doing some research into the last 3 tasks, I found that `resource.resource` is highly integrated into its own calendar model, `resource.calendar`. I need to integrate things a bit more with `calendar.event`, but was curious of your thoughts on which path I should take:

1. Create an exclusive calendar view for resources, and have it such that whenever a resource is added to a normal calendar event, the time slot of that event is directly mapped to the exclusive resource calendar.
2. Abandon the exclusive resource calendar view and fully integrate the resource leaves and resource availability times with `calendar.event`.

@lasley @dreispt

### DIFF
--- a/calendar_resource/README.rst
+++ b/calendar_resource/README.rst
@@ -10,11 +10,26 @@ This module adds new features to facilitate resource management with meetings:
 
 * To attach resources to a meeting, go to `Calendar` => Click on a meeting => `Edit` =>
   `Edit` => add any related resources.
-* To edit or add a resource, go to `Settings` => `Resource` => `Resources`.
+* To edit or add a resource, go to `Settings` => `Resource` => `Resources`. Developer mode
+  has to be enabled for this menu to show.
 * This module can also prevent resources from being double-booked. Go to a resource
   and disable `Allow Double Booking` (this is enabled by default). As a note, the same
-  resource in a meeting that ends at 5pm and also in a meeting that starts at 5pm on the same day
+  resource in a meeting that ends at 5pm and in a meeting that starts at 5pm on the same day
   would not not be considered double-booked.
+* You can restrict resources to certain calendar types. In the Resources form view, edit the
+  `Event Types` field. The resource can then be added to events only of those types.
+* Resources cannot be added to events when they're not available, either because they're on leave
+  or because the event time is outside their working time.
+* If you have an event selected as `allday`, only if there are no working times available
+  on 1 or more of the days in that event will an error be raised. For example, if you set
+  an event to `allday` on a Saturday and there are no working times at all that Saturday
+  for a particular resource, then an error will be raised. However, if there is at least
+  1 working time interval for that resource on the Saturday, regardless of how long or
+  short that working time is, the day will be considered covered by that working time.
+  If the event is not allday, any time the resource is not available during the event,
+  regardless of the time of day, will cause an error to be raised.
+* To stop leaves and working time validations on a resource when adding to an event,
+  set the resource's `Working Time` field to blank in Settings.
 
 Usage
 =====
@@ -23,14 +38,11 @@ Usage
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/111/10.0
 
-Known Issues / Roadmap
-======================
+Roadmap
+=======
 
-* Tie resources into calendar types (calendar.event.type), where certain resources are available
-  only with certain calendar types.
-* Add a calendar view on resources, so resources such as a room can be seen if available.
-* Validate Resource Leaves to avoid appointments for dates where the resource is not available.
-* Validate Resource Calendar, to avoid appointments for times outside their availability.
+* Abstract out logic from _check_resource_ids_working_times into separate methods in
+  resource.calendar.
 
 Bug Tracker
 ===========
--- a/calendar_resource/README.rst
+++ b/calendar_resource/README.rst
@@ -10,10 +10,9 @@ This module adds new features to facilitate resource management with meetings:
 
 * To attach resources to a meeting, go to `Calendar` => Click on a meeting => `Edit` =>
   `Edit` => add any related resources.
-* To edit or add a resource, go to `Settings` => `Resource` => `Resources`. Developer mode
-  has to be enabled for this menu to show.
+* To edit or add a resource, go to `Calendar` => `Resources` => `Resources`.
 * This module can also prevent resources from being double-booked. Go to a resource
-  and disable `Allow Double Booking` (this is enabled by default). As a note, the same
+  and disable or enable `Allow Double Booking` (this is disabled by default). As a note, the same
   resource in a meeting that ends at 5pm and in a meeting that starts at 5pm on the same day
   would not not be considered double-booked.
 * You can restrict resources to certain calendar types. In the Resources form view, edit the
@@ -26,10 +25,10 @@ This module adds new features to facilitate resource management with meetings:
   for a particular resource, then an error will be raised. However, if there is at least
   1 working time interval for that resource on the Saturday, regardless of how long or
   short that working time is, the day will be considered covered by that working time.
-  If the event is not allday, any time the resource is not available during the event,
+  If the event is not `allday`, any time the resource is not available during the event,
   regardless of the time of day, will cause an error to be raised.
 * To stop leaves and working time validations on a resource when adding to an event,
-  set the resource's `Working Time` field to blank in Settings.
+  set the resource's `Working Time` field to blank in the resource form view.
 
 Usage
 =====
--- a/calendar_resource/__manifest__.py
+++ b/calendar_resource/__manifest__.py
@@ -9,7 +9,7 @@
     "version": "10.0.1.1.0",
     "category": "CRM",
     "website": "http://www.savoirfairelinux.com",
-    "author": "Savoir-faire Linux, Laslabs, Odoo Community Association (OCA)",
+    "author": "Savoir-faire Linux, LasLabs, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "application": False,
     "installable": True,
--- a/calendar_resource/__manifest__.py
+++ b/calendar_resource/__manifest__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2010-2014 Savoir-faire Linux
-# Copryight 2017 Laslabs Inc.
+# Copyright 2017 Laslabs Inc.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
@@ -18,8 +18,11 @@
         "calendar",
     ],
     "data": [
+        "security/resource_security.xml",
+        "security/ir.model.access.csv",
         "views/calendar_event_view.xml",
         "views/resource_resource_view.xml",
+        "views/calendar_menu.xml",
     ],
     "demo": [
         "demo/resource_resource_demo.xml",
--- a/calendar_resource/__manifest__.py
+++ b/calendar_resource/__manifest__.py
@@ -1,11 +1,12 @@
 # -*- coding: utf-8 -*-
 # Copyright 2010-2014 Savoir-faire Linux
+# Copryight 2017 Laslabs Inc.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Calendar Resources",
     "summary": "New features to facilitate resource management with meetings.",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.1.0",
     "category": "CRM",
     "website": "http://www.savoirfairelinux.com",
     "author": "Savoir-faire Linux, Laslabs, Odoo Community Association (OCA)",
@@ -19,5 +20,10 @@
     "data": [
         "views/calendar_event_view.xml",
         "views/resource_resource_view.xml",
+    ],
+    "demo": [
+        "demo/resource_resource_demo.xml",
+        "demo/resource_calendar.xml",
+        "demo/resource_calendar_attendance.xml",
     ],
 }
--- a/calendar_resource/__manifest__.py
+++ b/calendar_resource/__manifest__.py
@@ -8,7 +8,7 @@
     "summary": "New features to facilitate resource management with meetings.",
     "version": "10.0.1.1.0",
     "category": "CRM",
-    "website": "http://www.savoirfairelinux.com",
+    "website": "https://github.com/OCA/crm",
     "author": "Savoir-faire Linux, LasLabs, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "application": False,
--- a/calendar_resource/demo/resource_calendar.xml
+++ b/calendar_resource/demo/resource_calendar.xml
@@ -5,7 +5,7 @@
 <odoo>
 
     <record id="resource_calendar_1" model="resource.calendar">
-        <field name="name">Calendar Resource Demo Calendar</field>
+        <field name="name">Resource Demo Calendar</field>
     </record>
 
 </odoo>
--- a/calendar_resource/demo/resource_calendar.xml
+++ b/calendar_resource/demo/resource_calendar.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 LasLabs Inc.
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record id="resource_calendar_1" model="resource.calendar">
+        <field name="name">Calendar Resource Demo Calendar</field>
+    </record>
+
+</odoo>
--- a/calendar_resource/models/__init__.py
+++ b/calendar_resource/models/__init__.py
@@ -2,5 +2,8 @@
 # Copyright 2014 Therp BV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from . import calendar_event_type
 from . import calendar_event
+from . import resource_calendar_attendance
+from . import resource_calendar
 from . import resource_resource
--- a/calendar_resource/models/__init__.py
+++ b/calendar_resource/models/__init__.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-# Copyright 2014 Therp BV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from . import calendar_event_type
--- a/calendar_resource/models/calendar_event.py
+++ b/calendar_resource/models/calendar_event.py
@@ -3,6 +3,8 @@
 # Copryight 2017 Laslabs Inc.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from datetime import datetime, time, timedelta
+
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
@@ -16,10 +18,88 @@ class CalendarEvent(models.Model):
         comodel_name='resource.resource',
     )
 
-    @api.constrains('resource_ids')
-    def _check_resource_ids_dbl_book(self):
+    @api.model
+    def _format_datetime_interval_list(self, intervals):
+        """ Converts intervals to string values for front-end
+            display purposes, taking into account language,
+            timezone, as well as date and time formats.
+
+        Example:
+
+            .. code-block python
+
+            self._format_datetime_interval_list(
+                intervals=[
+                    (
+                        datetime(2017, 6, 28, 17, 0, 0),
+                        datetime(2017, 6, 29, 8, 0, 0),
+                    ),
+                    (
+                        datetime(2017, 6, 29, 12, 0, 0),
+                        datetime(2017, 6, 29, 13, 0, 0),
+                    ),
+                ],
+            )
+
+            The above example will output the following string,
+            depending on date and time format, as well as timezone.
+
+            \n06/28/2017 at 17:00:00 To\n
+            06/29/2017 at 08:00:00 (UTC)\n
+
+            \n06/29/2017 at 12:00:00 To\n
+            06/29/2017 at 13:00:00 (UTC)\n
+
+        """
+        datetimes = ''
+        for interval in intervals:
+            if not isinstance(interval[0], str):
+                interval = (
+                    fields.Datetime.to_string(interval[0]),
+                    fields.Datetime.to_string(interval[1]),
+                )
+            args = {
+                'start': interval[0],
+                'stop': interval[1],
+                'zallday': False,
+                'zduration': 24,
+            }
+            datetimes += '\n%s\n' % self._get_display_time(**args)
+        return datetimes
+
+    @api.multi
+    def _event_in_past(self):
+        self.ensure_one()
+        stop_datetime = fields.Datetime.from_string(self.stop)
+        now_datetime = fields.Datetime.from_string(fields.Datetime.now())
+        return stop_datetime < now_datetime
+
+    @api.multi
+    def _get_event_date_list(self):
+        self.ensure_one()
+        start = fields.Date.from_string(self.start)
+        stop = fields.Datetime.from_string(self.stop)
+
+        if stop.time() == time(0, 0):
+            stop -= timedelta(days=1)
+
+        stop = stop.date()
+        date = start
+        dates = []
+        while date <= stop:
+            dates.append(date)
+            date += timedelta(days=1)
+
+        return dates
+
+    @api.multi
+    @api.constrains('resource_ids', 'start', 'stop')
+    def _check_resource_ids_double_book(self):
         """ Check Double Booking """
         for record in self:
+
+            if record._event_in_past():
+                continue
 
             resources = record.resource_ids.filtered(
                 lambda s: s.allow_double_book is False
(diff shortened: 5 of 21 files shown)


pedrobaeza:
I would go for integrating all (leaves, resources and so on) in the same calendar (calendar.event). It was the original intention of super_calendar module, and something like this is being done in Odoo master AFAIK.

woodbrettm:
Thanks @pedrobaeza for the feedback :). I'll move forward with fully integrating things as requested.

lasley:
@bmw95 - is this ready for review?

woodbrettm:
Woot. Everything is passing so ready for first round of review.

calendar_resource/__manifest__.py
```diff
@@ -1,11 +1,12 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Calendar Resources",
     "summary": "New features to facilitate resource management with meetings.",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.1.0",
     "category": "CRM",
     "website": "http://www.savoirfairelinux.com",
     "author": "Savoir-faire Linux, Laslabs, Odoo Community Association (OCA)",
```
lasley: s/Laslabs/LasLabs

calendar_resource/demo/resource_calendar.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 LasLabs Inc.
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record id="resource_calendar_1" model="resource.calendar">
+        <field name="name">Calendar Resource Demo Calendar</field>
```
lasley: Remove the first "Calendar" IMO

calendar_resource/models/calendar_event.py
```diff
@@ -16,11 +18,89 @@ class CalendarEvent(models.Model):
...
+            \n06/28/2017 at 17:00:00 To\n
+            06/29/2017 at 08:00:00 (UTC)\n
+
+            \n06/29/2017 at 12:00:00 To\n
+            06/29/2017 at 13:00:00 (UTC)\n
+
+        """
+        datetimes = ''
+        for interval in intervals:
+            if not isinstance(interval[0], str):
```
lasley: s/str/basestring

This will ensure Python 3 compat, which uses unicode internally IIRC

calendar_resource/models/calendar_event.py
```diff
@@ -16,11 +18,89 @@ class CalendarEvent(models.Model):
...
+            depending on date and time format, as well as timezone.
+
+            \n06/28/2017 at 17:00:00 To\n
+            06/29/2017 at 08:00:00 (UTC)\n
+
+            \n06/29/2017 at 12:00:00 To\n
+            06/29/2017 at 13:00:00 (UTC)\n
+
+        """
+        datetimes = ''
```
lasley: `datetimes = []`

calendar_resource/models/calendar_event.py
```diff
@@ -16,11 +18,89 @@ class CalendarEvent(models.Model):
...
+                    fields.Datetime.to_string(interval[0]),
+                    fields.Datetime.to_string(interval[1]),
+                )
+            args = {
+                'start': interval[0],
+                'stop': interval[1],
+                'zallday': False,
+                'zduration': 24,
+            }
+            datetimes += '\n%s\n' % self._get_display_time(**args)
```
lasley: `datetimes.append(`

calendar_resource/models/calendar_event.py
```diff
@@ -16,11 +18,89 @@ class CalendarEvent(models.Model):
...
+        return datetimes
```
lasley: `'\n'.join(datetimes)`

calendar_resource/models/calendar_event.py
```diff
@@ -16,11 +18,89 @@ class CalendarEvent(models.Model):
...
+
+    @api.multi
+    def _event_in_past(self):
+        self.ensure_one()
+        stop_datetime = fields.Datetime.from_string(self.stop)
+        now_datetime = fields.Datetime.from_string(fields.Datetime.now())
```
lasley: `now_datetime = datetime.now()`

calendar_resource/models/calendar_event.py
```diff
@@ -16,11 +18,89 @@ class CalendarEvent(models.Model):
...
+        return stop_datetime < now_datetime
+
+    @api.multi
+    def _get_event_date_list(self):
+        self.ensure_one()
+        start = fields.Date.from_string(self.start)
+        stop = fields.Datetime.from_string(self.stop)
+
+        if stop.time() == time(0, 0):
```
lasley: Is this midnight handling for overnight shifts?

calendar_resource/models/calendar_event.py
```diff
@@ -16,11 +18,89 @@ class CalendarEvent(models.Model):
...
+
+    @api.multi
+    def _event_in_past(self):
+        self.ensure_one()
+        stop_datetime = fields.Datetime.from_string(self.stop)
+        now_datetime = fields.Datetime.from_string(fields.Datetime.now())
+        return stop_datetime < now_datetime
+
+    @api.multi
+    def _get_event_date_list(self):
```
lasley: Am I understanding correctly that this method outputs a list of dates between two dates? If so, why not use `dateutil` instead of reinventing the wheel? Something like:
```
return list(
    rrule.rrule(rrule.DAILY, dtstart=start, until=end)
)
```

calendar_resource/models/calendar_event.py
```diff
@@ -34,12 +114,141 @@ def _check_resource_ids_dbl_book(self):
...
+    @api.multi
+    @api.constrains('resource_ids', 'start', 'stop')
+    def _check_resource_ids_leaves(self):
+
+        for record in self:
+
+            if record._event_in_past():
+                continue
+
+            for resource in record.resource_ids.filtered(
```
lasley: This is iterating the collection twice & doesn't really look that good. IMO it would be better to just 
```
if not s.calendar_id or not s.calendar.leave_ids: 
    continue
```
instead of filter

calendar_resource/models/calendar_event.py
```diff
@@ -34,12 +114,141 @@ def _check_resource_ids_dbl_book(self):
...
+
+                    if not intervals and record.allday:
+                        conflict_intervals.append(
+                            (datetime_start, datetime_end)
+                        )
+                    else:
+                        available_intervals += intervals
+
+                if not record.allday:
+                    conflict_intervals = self.env['resource.calendar'].\
```
lasley: Missing a test for this

calendar_resource/models/calendar_event.py
```diff
@@ -34,12 +114,141 @@ def _check_resource_ids_dbl_book(self):
...
+                            available_intervals, event_start, event_stop
+                        )
+
+                if not conflict_intervals:
+                    continue
+
+                if record.allday:
+                    conflict_intervals = self.env['resource.calendar'].\
+                        _remove_datetime_interval_overlaps(
+                            conflict_intervals
```
lasley: trailing comma

calendar_resource/models/calendar_event.py
```diff
@@ -34,12 +114,141 @@ def _check_resource_ids_dbl_book(self):
...
+                raise ValidationError(
+                    _(
+                        'The resource, %s, is not available during '
+                        'the following dates and times which are '
+                        'conflicting with the event:\n%s'
+                    )
+                    % (
+                        resource.name,
+                        self._format_datetime_interval_list(
+                            conflict_intervals
```
lasley: trailing comma

calendar_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class ResourceCalendar(models.Model):
+
+    _inherit = 'resource.calendar'
+
+    @api.model
+    def _get_conflicting_unavailable_intervals(self, intervals,
+                                               start_datetime, end_datetime):
+        """ Finds all unavailable datetime gaps in intervals argument that
```
lasley: Take a look at [PEP-257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings). Your first line of the docblock can't continue into another. Find a way to shorten this description, and optionally add the longer supplementary explanation below first line.

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

calendar_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,165 @@
...
+                    .. code-block:: python
+
+                    intervals = \
+                        resource.calendar_id.get_working_intervals_of_day(
+                            start_dt=datetime_start,
+                            end_dt=datetime_end,
+                            resource_id=resource.id,
+                        )
+
+            start_datetime, end_datetime (datetime): Datetime values
```
lasley: Each arg needs its own line. https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

calendar_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,165 @@
...
+        return conflicting_intervals
+
+    @api.model
+    def _get_unavailable_intervals(self, intervals, start_date, end_date):
+        """ Finds any gaps between intervals, the beginning of start_date,
+            and the end of end_date.
+
+        """
+        start_datetime = datetime.combine(
+            start_date, time(00, 00, 00)
```
lasley: trailing comma

calendar_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,165 @@
...
+        )
+        end_datetime = datetime.combine(
+            end_date, time(23, 59, 59)
```
lasley: trailing comma

calendar_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,165 @@
...
+        )
+
+        intervals = self._remove_datetime_interval_overlaps(intervals)
+
+        unavailable_intervals = []
+
+        if intervals[0][0] > start_datetime:
+            unavailable_intervals.append(
+                (start_datetime, intervals[0][0])
```
lasley: trailing comma

calendar_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,165 @@
...
+            )
+
+        if intervals[-1][1] < end_datetime:
+            unavailable_intervals.append(
+                (intervals[-1][1], end_datetime)
```
lasley: trailing comma

calendar_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,165 @@
...
+
+            if all((interval[0] < end_datetime,
+                    interval[1] > start_datetime)):
+                conflicting_intervals.append(interval)
+
+        return conflicting_intervals
+
+    @api.model
+    def _get_unavailable_intervals(self, intervals, start_date, end_date):
+        """ Finds any gaps between intervals, the beginning of start_date,
```
lasley: Single line.

Also, I'm having a lot of trouble keeping these data structures straight. What is intervals?

calendar_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,165 @@
...
+                (start_datetime, intervals[0][0])
+            )
+
+        if intervals[-1][1] < end_datetime:
+            unavailable_intervals.append(
+                (intervals[-1][1], end_datetime)
+            )
+
+        if len(intervals) < 2:
+            return unavailable_intervals
```
lasley: Untested

calendar_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,165 @@
...
+
+        for index, dt_interval in enumerate(intervals):
+
+            if index - 1 < 0:
+                continue
+
+            previous_pair = intervals[index - 1]
+            if previous_pair[1] < dt_interval[0]:
+                unavailable_intervals.append(
+                    (previous_pair[1], dt_interval[0])
```
lasley: trailing comma

calendar_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,165 @@
...
+                )
+
+        return self._check_round_up_times_to_next_day(
+            sorted(unavailable_intervals, key=lambda s: s[0])
```
lasley: trailing comma

calendar_resource/models/resource_calendar.py
```diff
@@ -0,0 +1,165 @@
...
+        )
+
+    @api.model
+    def _remove_datetime_interval_overlaps(self, intervals):
+
+        intervals = self._check_round_up_times_to_next_day(
+            sorted(intervals, key=lambda s: s[0])
```
lasley: trailing comma

lasley (changes requested):

calendar_resource/models/calendar_event.py
woodbrettm: This is correct as a string, however I think the method should be renamed as it's misleading about what's being returned.

calendar_resource/demo/resource_calendar.xml
woodbrettm: lol "Calendar Calendar Calendar Resource Calendar Calendar".

calendar_resource/models/calendar_event.py
woodbrettm: oh nice. Didn't realize that.

calendar_resource/models/calendar_event.py
woodbrettm: yeah I was debating about that. Looks cleaner with the check.

calendar_resource/tests/setup.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+
+        return (start_datetime, end_datetime)
+
+    def _create_event(self, vals=None):
+        create_vals = {
+            'name': 'Test Event',
+            'resource_ids': [(6, 0, [self.resource_1.id])],
+            'categ_ids': [(6, 0, [self.event_type_4.id])],
+            'allday': True,
```
woodbrettm, line 165: This should be set to False and tests I want to be allday should be overriding instead.

calendar_resource/models/resource_calendar.py
woodbrettm: hmm. I think in the docstring I'll put an example what the interval arg would look like. I'm using a method in the `resource` module that returns all the working times for the day in the following format:
`[(start_datetime, stop_datetime), (start_datetime, stop_datetime)]`. So basically a list of all the intervals available.

calendar_resource/models/calendar_event.py
woodbrettm: Thanks for mentioning. Not being tested pointed to a detail I missed in tests which needs to be fixed.

calendar_resource/models/resource_calendar.py
woodbrettm: ah. I overlooked that while reading.

calendar_resource/models/calendar_event.py
woodbrettm: No all of that is done in the remove_interval_time_overlaps method and the round_up method in resource.calendar. 

This just ensures that if the stop time ends at the beginning of a day, not to include that day (which I now realize would make this method highly misleading). When I loop through all the days of the event, if I included that day then all the working times for that day would be unnecessarily retrieved.

I originally tried working with rrule, but found there to be too many edge cases for me to work with it, so I built this custom method.

Now that the edge cases have been ironed out, it seems I could just forego this custom method altogether now, go back to rrule, and simply remove the last date from the list if this edge case is true.

calendar_resource/models/resource_calendar.py
woodbrettm: Wow looks like I missed a lot of these. Thanks for catching.

calendar_resource/models/calendar_event.py
woodbrettm: I ended up not using rrule, as I found if the difference between the stop and start datetimes is less than 24h, (e.g. monday 9am to tuesday, 1am), rrule would not include tuesday. Perhaps there's a way around that, but I left my current custom _get_event_date_list in for now.

calendar_resource/tests/setup.py
woodbrettm: Changed the tests to specifically define allday in each test for better clarity.

woodbrettm:
Ready for next round of review.

calendar_resource/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     "version": "10.0.1.1.0",
     "category": "CRM",
     "website": "http://www.savoirfairelinux.com",
```
lasley: Let's link to the repo here

calendar_resource/models/calendar_event.py
```diff
@@ -64,38 +68,20 @@ def _format_datetime_interval_list(self, intervals):
                 'zallday': False,
                 'zduration': 24,
             }
-            datetimes += '\n%s\n' % self._get_display_time(**args)
-        return datetimes
+            datetimes.append(self._get_display_time(**args))
+        return '\n\n'.join(datetimes)
```
lasley: This won't work as expected on Windows [*shudder*] servers. Use `os.linesep` 

```
>>> import os
>>> os.linesep
'\n'
```

calendar_resource/models/calendar_event.py
```diff
@@ -178,15 +169,33 @@ def _check_resource_ids_leaves(self):
...
+        start = fields.Date.from_string(self.start)
+        stop = fields.Datetime.from_string(self.stop)
+
+        if stop.time() == time(0, 0):
+            stop -= timedelta(days=1)
+
+        stop = stop.date()
+        date = start
+        dates = []
+        while date <= stop:
```
lasley: IMO it's best to avoid `while` and `until` loops whenever possible - they can cause infinite loops because there is no explicit ending. The main time I make an exception is something like:

```
arr = [1, 2, 3]
while len(arr):
  val = arr.pop()
  if val % 2:
    arr.append(val * 2)
```

In the above, we technically have an explicit ending because the list is sure to end at some point. We also couldn't use the standard range iteration tactic, because the list is being mutated from underneath the loop. Neither of those conditions apply here.

Maybe use a [`dateutil.rrule`](https://dateutil.readthedocs.io/en/stable/rrule.html) instead? Or just something like [this](https://stackoverflow.com/questions/1060279/iterating-through-a-range-of-dates-in-python)?

calendar_resource/models/calendar_event.py
```diff
@@ -219,36 +229,36 @@ def _check_resource_ids_working_times(self):
...
                     if not intervals and record.allday:
                         conflict_intervals.append(
-                            (datetime_start, datetime_end)
+                            (datetime_start, datetime_end),
                         )
                     else:
                         available_intervals += intervals
 
                 if not record.allday:
                     conflict_intervals = self.env['resource.calendar'].\
```
lasley: This `\` is killing me. Let's assign the model to an intermediary variable or something instead

calendar_resource/models/calendar_event.py
```diff
@@ -219,36 +229,36 @@ def _check_resource_ids_working_times(self):
...
                         _get_conflicting_unavailable_intervals(
-                            available_intervals, event_start, event_stop
+                            available_intervals, event_start, event_stop,
                         )
 
                 if not conflict_intervals:
                     continue
 
                 if record.allday:
                     conflict_intervals = self.env['resource.calendar'].\
```
lasley: Same thing re `\`. It's a dirty hack - don't do it.

calendar_resource/models/calendar_event.py
```diff
@@ -219,36 +229,36 @@ def _check_resource_ids_working_times(self):
...
+                        _clean_datetime_intervals(
+                            conflict_intervals,
                         )
 
                 raise ValidationError(
                     _(
                         'The resource, %s, is not available during '
                         'the following dates and times which are '
-                        'conflicting with the event:\n%s'
+                        'conflicting with the event:\n\n%s',
```
lasley: `os.linesep`

lasley (changes requested):

calendar_resource/models/calendar_event.py
woodbrettm: hmm is it `.join` that doesn't work as expected, or is it using the double newlines that acts weird?

lasley: The issue is the newline char itself, which differs between operating systems - https://en.wikipedia.org/wiki/Newline

On your Windows box, newline is `\r\n`

woodbrettm: sigh . . . Microsoft at it again.

lasley: History lesson - before OSX (which is Unix based), Mac line endings were `\r`

woodbrettm: ah

woodbrettm:
@lasley comments attended to

calendar_resource/README.rst
```diff
@@ -10,11 +10,26 @@ This module adds new features to facilitate resource management with meetings:
 
 * To attach resources to a meeting, go to `Calendar` => Click on a meeting => `Edit` =>
   `Edit` => add any related resources.
-* To edit or add a resource, go to `Settings` => `Resource` => `Resources`.
+* To edit or add a resource, go to `Settings` => `Resource` => `Resources`. Developer mode
+  has to be enabled for this menu to show.
```
dreispt: Nice to have: add a new a menu item fro resources somewhere else, to not require the Developer Mode.

calendar_resource/README.rst
```diff
@@ -10,11 +10,26 @@ This module adds new features to facilitate resource management with meetings:
...
+* If you have an event selected as `allday`, only if there are no working times available
+  on 1 or more of the days in that event will an error be raised. For example, if you set
+  an event to `allday` on a Saturday and there are no working times at all that Saturday
+  for a particular resource, then an error will be raised. However, if there is at least
+  1 working time interval for that resource on the Saturday, regardless of how long or
+  short that working time is, the day will be considered covered by that working time.
+  If the event is not allday, any time the resource is not available during the event,
+  regardless of the time of day, will cause an error to be raised.
+* To stop leaves and working time validations on a resource when adding to an event,
+  set the resource's `Working Time` field to blank in Settings.
```
dreispt: +1

calendar_resource/models/__init__.py
```diff
@@ -2,5 +2,8 @@
 # Copyright 2014 Therp BV
```
dreispt: Curiously, Therp in not in the manifest Authors list...

dreispt (approved):
Made a light code review, and it LGTM

calendar_resource/README.rst
woodbrettm: Yeah I'm not sure exactly why it was placed there originally in resource. Perhaps we should create a menu in the top navbar? Name it 'Resources'.

calendar_resource/models/__init__.py
lasley: This is the only Therp I see too. Perhaps someone just copied an init file and forgot to change copyright?

calendar_resource/README.rst
dreispt: This module depend on the Calendar app. So one idea is to add it there.
Maybe we could move the "Settings > Technical > Calendar" menu to "Calendar > Configuration", not forgetting to remove the "Technical Features" group from it's Access Rights, so that other people can see it. 
This new "resources" menu should go there.
And I think we need a new "Resource manager" user to control access to these configurations...

Just bringing more work, sorry... 😢

calendar_resource/models/__init__.py
dreispt: I bet so. It should be OK to remove it.

calendar_resource/README.rst
woodbrettm: @dreispt You mean adding to the Calendar page directly? I see there's some space on the right there, but I don't think I've seen any apps actually populate that area before. Perhaps there are and I've just never seen any. Is that where you're thinking?
fireshot capture 8 - meetings - odoo_ - http___192 168 1 133_8069_web

lasley: @bmw95 - not in the page directly, but just in the top menu. Although it looks like there's already a `Settings` though so I think we'd create confusion with the proposed `Configuration` header. What about `Resources` then your two menu items as sub items?

woodbrettm: That's what I was thinking yeah. Especially since the concept of Resources is becoming more prominent. Should I hide the resources menus in the technical settings then, or just leave them as is?

lasley: The advice was to move your current menus into the calendar menu. What other menus are you referring to?

woodbrettm: Aight so the plan is to create a new header in the top bar called `Resources` right? There are resource views in technical settings, so I need to move those to the new `Resources` menu.

That's what I'm getting at least. But I'm not sure whether to leave the resources menus in technical settings as-is or remove them.

lasley: Ohhhhhhh I see - sorry I didn't realize there were already existing settings there & you were adding to the menu that existed. Ummm let's move it all under Calendar as a Resources subheader, which sounds like exactly what you're understanding too I think?

woodbrettm: I think so. These menus in the picture, from `Settings > Technical` will be moved under `Resources > Calendar` 

cools

Perhaps only working time and resource leaves should go under `Resources > Calendar` and then that resources menu in the picture should go under `Resources > Resources`

Should I remove the existing menus from technical settings though?

dreispt: Maybe just add a couple (more if needed) of menu items under the "Calendar" top menu.
Here is a mockup:

2017-08-11 11_35_04-meetings - odoo


I'm not sure it is relevant enough to move the "Settings/Technical/Calendar" options to "Calendar/Configuration".

lasley: Hmmm it seems we have a problem here - the Calendar in v10 doesn't have a sidebar so it's not possible to add menu items:

image

Not really sure the best place in this instance 😦

dreispt: My screenshot is from v10. 
The sidebar shows up as son as you add submenus.

woodbrettm: @dreispt are there any other modules that have done this? I feel like there's an unspoken rule not to put anything there or attempt to mess with that page too much.

lasley: I'm good with the sidebar automatically appearing. We don't have a sidebar in our UI anyways

dreispt: > We don't have a sidebar in our UI anyways

I didn't get that. The sidebar menu is the standard way menus are presented in CE.

lasley: [`web_responsive`](https://github.com/OCA/web/tree/10.0/web_responsive) is our preferred CE UI, which doesn't have a side bar

woodbrettm:
Aight updated! Sorry for the delay I found a few bugs that had to be fixed. 

In the latest commit, a new security group for resource_managers has been created, and the resource menus are moved over to underneath the calendar header.

lasley (approved):
LGTM thanks @bmw95

PR #150, opened 2017-05-25
[10.0] Adjustment, remove modules which depend on crm_claim (not exist on odoo 10)

bimajatiwijaya:
- No module here depend on crm_claim_code and crm_claim_type
- Module crm_claim is deleted on odoo 10.0
- Both modules will move to OCA/rma. All needed object from crm_claim odoo9 will move to crm_claim_rma which located on repo OCA/rma and both module will depend on it.


rafaelbn (changes requested):
There is not a decision about what to do about crm_claim, an Odoo module in v9 which disappear in v10. So until a decision we cannot delete anything.

bimajatiwijaya:
Thank you for your review @rafaelbn .
Next time i will report issue before create a pull request.

rafaelbn:
OK @bimajatiwijaya then I close by the moment

PR #157, opened 2017-06-08
[10.0][MIG] crm_location

luismontalba:
CRM location
-----------------
This module introduces a better zip.
It enables zip, city, state and country auto-completion on lead.

cc @Tecnativa

### DIFF
--- a/crm_location/__manifest__.py
+++ b/crm_location/__manifest__.py
@@ -1,39 +1,20 @@
 # -*- coding: utf-8 -*-
-# Python source code encoding : https://www.python.org/dev/peps/pep-0263/
-##############################################################################
-#
-#    OpenERP, Open Source Management Solution
-#    This module copyright :
-#        (c) 2015 Antiun Ingenieria, SL (Madrid, Spain, http://www.antiun.com)
-#                 Endika Iglesias <endikaig@antiun.com>
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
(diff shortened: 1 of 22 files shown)


luismontalba:
Travis is failing due to this module depends on base_location located in OCA/partner-contact repo.

pedrobaeza:
It shouldn't fail, because you can specify on oca_repositories.txt these kind of dependencies.

crm_location/models/crm_lead.py
```diff
@@ -45,11 +26,8 @@ def on_change_city(self):
     )
 
     @api.multi
-    def on_change_partner_id(self, partner_id):
-        res = super(CrmLead, self).on_change_partner_id(partner_id)
-        if 'value' not in res:
-            res['value'] = {}
+    def _onchange_partner_id_values(self, partner_id):
```
pedrobaeza: This old onchange doesn't work for sure. Please use new API onchange (you can create a new method for not having to overwrite super):
```
@api.onchange('partner_id')
def onchange_partner_id_crm_location(self):
    if self.partner_id:
        self.location_id = self.partner_id.zip_id.id
```

crm_location/tests/test_crm_location.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertEqual(lead.state_id, location.state_id)
+        self.assertEqual(lead.country_id, location.country_id)
+
+    def test_onchange_partner_id_values(self):
+        lead = self.lead
+        partner = self.partner
+        location = self.location
+        partner.zip_id = location.id
+        lead.partner_id = partner.id
+        lead._onchange_partner_id_values(lead.partner_id.id)
```
pedrobaeza: Call the new onchange method.

pedrobaeza (changes requested):
And squash your commits together

luismontalba:
Changes proposed done and own commits squashed.

crm_location/__manifest__.py
```diff
@@ -1,39 +1,20 @@
...
     'category': 'Customer Relationship Management',
-    'version': '8.0.1.0.0',
+    'version': '10.0.1.0.0',
     'depends': [
         'crm',
         'base_location',
     ],
     'data': ['views/crm_lead_view.xml'],
-    'author': 'Antiun Ingeniería S.L., '
+    'author': 'Antiun Ingeniería, '
```
yajo: Add Tecnativa

yajo (approved):

crm_location/__manifest__.py
pedrobaeza: Better to replace, as this version is not maintained anymore by these old companies

luismontalba:
Author replaced, new commit squashed.

pedrobaeza (approved):

PR #159, opened 2017-06-08
[10.0][MIG] crm_deduplicate_filter

luismontalba:
Exclude some records from the deduplication
---------------------------------------------------------
This module extends the possibilities of the contact deduplication allowing
to filter the applicable set according to several criteria.

For now, only the filter for restricting the deduplication to only companies or
only contacts is implemented.

Usage
--------

* Go to *CRM/Sales > Tools > Deduplicate Contacts*.
* Mark "'Is a company?' field selected", "Parent company not set" or
   "Parent company set (Contacts)" in the section 'Exclude contacts having'.
* This criteria will be used for excluding in the deduplication the selected
   kind of records.

cc @Tecnativa


pedrobaeza (approved):
Code review OK
Functional test OK:
* Created 'Test' company partner.
* Created 'Test' individual partner.
* Run deduplication wizard with name option and without filters. It proposes to merge both.
* Run deduplication wizard with name option and filter `'Is a company?' field selected`

There's only a comment:

* Now that the fields have been renamed, you can update the label of the first filter to "Company partners (not individuals)" or similar.

pedrobaeza:
@rafaelbn please review

rafaelbn (approved):
Tested functionally in runbot 👍 
I work fine, maybe is complicate to understand how deduplication work in Odoo core.... 😄 . For example: when you click 2 checkbox it applies AND and not OR

crm_deduplicate_filter/wizards/partner_merge.py
```diff
@@ -0,0 +1,35 @@
...
+class BasePartnerMergeAutomaticWizard(models.TransientModel):
+    _inherit = "base.partner.merge.automatic.wizard"
+
+    exclude_is_company = fields.Boolean("'Is a company?' field selected")
+    exclude_not_parent = fields.Boolean("Parent company not set")
+    exclude_parent = fields.Boolean("Parent company set (Contacts)")
+
+    @api.multi
+    def _process_query(self, query):
+        if any([self.exclude_is_company, self.exclude_not_parent,
```
yajo, line 17: I think this `if` is a bit redundant, since it makes the individual `if`s just below.
I'd do all 3 individual `if`s and after that a single `if filters:`.
It's not a blocker though, since the final result would be the same.

yajo (approved):
Code review.

crm_deduplicate_filter/wizards/partner_merge.py
pedrobaeza: This is only for performance. Avoid decomposing the query if there are no filters.

PR #162, opened 2017-06-16
[FIX] crm_lead_vat: fix onchange error

oihane:
capturadepantallade2017-06-16115715

This is the fixed error.

### DIFF
--- a/crm_lead_vat/models.py
+++ b/crm_lead_vat/models.py
@@ -22,9 +22,10 @@ class Lead(models.Model):
     def on_change_partner_id(self, partner_id):
         """Recover VAT from partner if available."""
         result = super(Lead, self).on_change_partner_id(partner_id)
-
-        if result.get("value"):
-            if self.partner_id and self.partner_id.vat:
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

agaldona (approved):

crm_lead_vat/models.py
```diff
@@ -22,9 +22,10 @@ def _lead_create_contact(self, lead, name, is_company, parent_id=False):
     def on_change_partner_id(self, partner_id):
```
yajo, line 22: Although the fix is good, I feel the real fix to your error would be adding `@api.multi` for this method...

yajo (approved):

yajo:
Check the bots please.

crm_lead_vat/models.py
oihane: That was the first option, but as it is an old onchange it does not work as expected. So this was the real fix.

oihane:
@Yajo it's already green

PR #165, opened 2017-07-20
[9.0][MIG] crm_helpdesk

MiquelRForgeFlow:
Migrated from Odoo, 8.0 version.


crm_helpdesk/data/crm_helpdesk_data.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<openerp>
+    <data noupdate="1">
+    <!--
+    Case Sections
+    -->
+    </data>
```
LoisRForgeFlow: Why this file?

LoisRForgeFlow (approved):
Tested on runbot :+1:

crm_helpdesk/data/crm_helpdesk_data.xml
MiquelRForgeFlow: This file existed in v8 😂

LoisRForgeFlow: I see... a pity to remove it, the nostalgia is very powerful :laughing: I would do it though!

MiquelRForgeFlow: now it's removed

PR #166, opened 2017-07-20
[10.0][MIG] crm_helpdesk

MiquelRForgeFlow:
Standard migration from https://github.com/OCA/crm/pull/165.


crm_helpdesk/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+# Copyright 2017 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'CRM Helpdesk',
+    'category': 'Customer Relationship Management',
+    'version': '10.0.1.0.0',
+    'summary': "Helpdesk Management",
+    'author': 'OpenERP SA, Eficent, '
+              'Odoo Community Association (OCA)',
+    'website': 'https://github.com/Eficent/ao-odoo',
```
LoisRForgeFlow: change this.

crm_helpdesk/tests/test_crm_helpdesk.py
```diff
@@ -0,0 +1,63 @@
...
+            sudo(self.helpdesk_user).search(
+            [('email_from', '=', 'Mr. John Right <info@customer.com>')],
+            limit=1)
+        try:
+            self.env['crm.helpdesk'].\
+                sudo(self.helpdesk_user).message_update(
+                question,
+                {'subject': 'Link of product', 'body': 'www.odoo.com'})
+        except Exception:
+            pass
```
LoisRForgeFlow: What is the purpose of this try-except block?

LoisRForgeFlow (commented):
Tested locally, It works fine. Just a pair of comments

crm_helpdesk/tests/test_crm_helpdesk.py
MiquelRForgeFlow: Don't know, it was from odoo v8 <code>help-desk.yml</code> file

LoisRForgeFlow: In any case it is incorrect, it comes from https://github.com/odoo/odoo/blob/8.0/addons/crm_helpdesk/test/process/help-desk.yml#L25. So it should be: `question.message_update({'subject': 'Link of product', 'body': 'www.odoo.com'})` in the new api. And then instead of `pass` I would `raise` to have the tests failing if something goes wrong.

MiquelRForgeFlow: No. See https://github.com/OCA/OCB/blob/10.0/addons/mail/models/mail_thread.py#L1322

LoisRForgeFlow: look at the method you refer:

`def message_update(self, msg_dict, update_vals=None):`

you are calling it here like this:

`(...).message_update(question, {'subject': 'Link of product', 'body': 'www.odoo.com'})`

there, `question` is [the result of a search on helpdesk model](https://github.com/OCA/crm/pull/166/files#diff-905bd1d8172234ad0a77d084c9e66c30R53) and that dictionary is what is expected for `msg_dict`, not for `update_vals`

So, I recall, it should be:
`question.message_update({'subject': 'Link of product', 'body': 'www.odoo.com'})`

MiquelRForgeFlow: Done

LoisRForgeFlow (approved):
Great, Thanks! LGTM now :+1:

PR #169, opened 2017-09-04
[FIX] crm_phonecall: wrong oca repo id

cubells:
We can't test module if not :smile: 

cc @Tecnativa


pedrobaeza (approved):
Fast-tracking this

PR #173, opened 2017-09-18
[IMP][crm_phonecall] Add utm.mixin fields

yajo:
This should let a user select campaign, medium and source for phonecalls, allowing better tracing for them.

Also those fields land in the converted opportunity or rescheduled call.

@Tecnativa

### DIFF
--- a/crm_phonecall/views/crm_phonecall_view.xml
+++ b/crm_phonecall/views/crm_phonecall_view.xml
@@ -100,7 +100,7 @@
                         <field name="priority" widget="priority"/>
                         <field name="opportunity_id" context="{'opportunity_id': opportunity_id}"/>
                     </group>
-                    <group string="Tracking" groups="base.group_multi_company,base.group_no_one" name="categorization">
+                    <group string="Tracking" groups="base.group_no_one" name="categorization">
                         <field name="campaign_id"/>
                         <field name="source_id"/>
                         <field name="medium_id"/>
@@ -130,6 +130,9 @@
                 <field name="state" invisible="1"/>
                 <field name="create_date" invisible="1"/>
                 <field name="opportunity_id" invisible="1"/>
+                <field name="campaign_id" groups="base.group_no_one"/>
+                <field name="source_id" groups="base.group_no_one"/>
+                <field name="medium_id" groups="base.group_no_one"/>
                 <button string="Schedule Other Call"
                     icon="terp-call-start"
                     name="%(phonecall_to_phonecall_act)d"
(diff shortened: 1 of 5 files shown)


pedrobaeza:
@chienandalu @cubells please review

cubells (approved):
:+1: 

Code review and tests.

crm_phonecall/views/crm_phonecall_view.xml
```diff
@@ -100,6 +100,11 @@
                         <field name="priority" widget="priority"/>
                         <field name="opportunity_id" context="{'opportunity_id': opportunity_id}"/>
                     </group>
+                    <group string="Tracking" groups="base.group_multi_company,base.group_no_one" name="categorization">
+                        <field name="campaign_id"/>
+                        <field name="source_id"/>
+                        <field name="medium_id"/>
```
chienandalu: Shouldn't this fields also be available to record at the logged calls editable tree?

yajo: Mmm I thought about that, I'm not sure. Some users could consider that excessive information, others necessary... @pedrobaeza @rafaelbn help here please.

pedrobaeza: Isn't there any specific security group for this?

crm_phonecall/views/crm_phonecall_view.xml
```diff
@@ -100,6 +100,11 @@
                         <field name="priority" widget="priority"/>
                         <field name="opportunity_id" context="{'opportunity_id': opportunity_id}"/>
                     </group>
+                    <group string="Tracking" groups="base.group_multi_company,base.group_no_one" name="categorization">
```
pedrobaeza: Why multi-company here?

yajo: I copied it from [the crm.lead view](https://github.com/odoo/odoo/blob/c14eb9ae9e29977be27bc0370d7662a6acd6ab9a/addons/crm/crm_lead_view.xml#L181), thinking that it would be confusing for a user to see the fields in crm.phonecall and not in crm.lead.

pedrobaeza: That one includes the field `company_id`, so that's the reason. That's a bug clearly as you can use tracking info without multi-company. Please remove it here and optionally make the PR in Odoo

yajo: Then do I leave the `base.group_no_one` in place? And do I add the fields with the same group to the tree view?

pedrobaeza: Yes, you can leave that one considering these fields as "technical".

yajo:
Fix for Odoo in https://github.com/odoo/odoo/pull/19551, multicompany group removed, fields added to tree view.

pedrobaeza (approved):

chienandalu (approved):

PR #174, opened 2017-09-20
[ADD][crm_phonecall_planner] Phonecall autoscheduler

yajo:
A simple to use wizard that generates phonecall plannings under given criteria.

See README for further details.

WIP until:

- [x] https://github.com/OCA/partner-contact/pull/475

@Tecnativa


crm_phonecall_planner/README.rst
```diff
@@ -0,0 +1,120 @@
...
+Phonecall planner
+=================
+
+Use this module to plan a phone calls schedule for your partners, assuming you
+have specified the ideal time to call them.
+
+Usage
+=====
+
+To use this module, you need to specify your partners' preferred phone call
```
pedrobaeza: Put this paragraph on "Configuration" section.

crm_phonecall_planner/README.rst
```diff
@@ -0,0 +1,120 @@
...
+------------
+
+* Jairo Llopis <jairo.llopis@tecnativa.com>
+
+Do not contact contributors directly about support or help with technical issues.
+
+Funders
+-------
+
+The development of this module has been financially supported by:
```
pedrobaeza: Remove this as it's not actually true (this is for final customers).

crm_phonecall_planner/README.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Jairo Llopis <jairo.llopis@tecnativa.com>
```
pedrobaeza: Put here Tecnativa URL

crm_phonecall_planner/README.rst
pedrobaeza: Maybe we can propose a change in the template for grouping contributors by company, putting something like:
```rst
* Tecnativa (https://www.tecnativa.com)
  * Jairo Llopis <option e-mail>
...
```

crm_phonecall_planner/tests/test_planner.py
```diff
@@ -0,0 +1,118 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Jairo Llopis <jairo.llopis@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp.tests.common import SavepointCase
+
+
+class SomethingCase(SavepointCase):
```
pedrobaeza: SomethingCase?

crm_phonecall_planner/wizards/crm_phonecall_planner.py
```diff
@@ -0,0 +1,212 @@
...
+        comodel_name="crm.lead.tag",
+        string="Tags",
+    )
+    res_partner_domain = fields.Char(
+        "Partners filter",
+        help="Filter the parters that will get a scheduled call.",
+    )
+    duration = fields.Float(
+        string="Call duration",
+        default=lambda self: self._default_duration(),
```
pedrobaeza, line 44: Don't do this, but move to up the default method and just put `default=_default_duration` (without quotes in the method).

crm_phonecall_planner/wizards/crm_phonecall_planner.py
```diff
@@ -0,0 +1,212 @@
...
+        required=True,
+        help="Leave this free time between phone calls.",
+    )
+    start = fields.Datetime(
+        default=lambda self: self._default_start(),
```
pedrobaeza: The same

crm_phonecall_planner/wizards/crm_phonecall_planner.py
```diff
@@ -0,0 +1,212 @@
...
+        required=True,
+        help="Schedule calls from this moment. The time you select will be "
+             "used as the plan starting time for each day in the range.",
+    )
+    end = fields.Datetime(
+        default=lambda self: self._default_end(),
```
pedrobaeza: The same

crm_phonecall_planner/wizards/crm_phonecall_planner.py
```diff
@@ -0,0 +1,212 @@
...
+    def _default_start(self):
+        return datetime.now()
+
+    @api.model
+    def _default_end(self):
+        return self._default_start() + timedelta(days=30, hours=8)
+
+    @api.constrains("start", "end")
+    def _constrains_plan_dates(self):
+        if self.start > self.end:
```
pedrobaeza: You have to make it multi-record aware

crm_phonecall_planner/wizards/crm_phonecall_planner.py
```diff
@@ -0,0 +1,212 @@
...
+            "name": "Generated calls",
+            "type": "ir.actions.act_window",
+            "res_model": "crm.phonecall",
+            "views": [[False, "tree"], [False, "calendar"], [False, "form"]],
+            "domain": [("id", "in", self.planned_calls.ids)],
+        }
+
+    @api.multi
+    def _schedule_call(self, partner, when):
+        call = self.env["crm.phonecall"]
```
pedrobaeza: To reduce cyclomatic complexity, do:
```python
if not partner:
    return self.env["crm.phonecall"]
_logger.debug(
...
```

pedrobaeza (changes requested):

crm_phonecall_planner/wizards/crm_phonecall_planner.py
yajo: Nope, check the guidelines: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#fields

(same in comments below)

crm_phonecall_planner/wizards/crm_phonecall_planner.py
yajo: I think I'll just remove the returned value, it's not used after all.

yajo:
This is testable now.

rafaelbn:
Hi @Yajo ,

In this use case I get an error:



```` 
Odoo Server Error

  File "/.repo_requirements/odoo/openerp/http.py", line 650, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/.repo_requirements/odoo/openerp/http.py", line 687, in dispatch
    result = self._call_function(**self.params)
  File "/.repo_requirements/odoo/openerp/http.py", line 323, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/service/model.py", line 118, in wrapper
    return f(dbname, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/http.py", line 316, in checked_call
    result = self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/openerp/http.py", line 966, in __call__
    return self.method(*args, **kw)
  File "/.repo_requirements/odoo/openerp/http.py", line 516, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo-9.0/addons/web/controllers/main.py", line 899, in call_button
    action = self._call_kw(model, method, args, {})
  File "/home/odoo/odoo-9.0/addons/web/controllers/main.py", line 887, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/api.py", line 250, in wrapper
    return old_api(self, *args, **kwargs)
  File "/.repo_requirements/odoo/openerp/api.py", line 381, in old_api
    result = method(recs, *args, **kwargs)
  File "/home/odoo/build/OCA/crm/crm_phonecall_planner/wizards/crm_phonecall_planner.py", line 128, in action_accept
    partner_domain = safe_eval(self.res_partner_domain) + [
  File "/.repo_requirements/odoo/openerp/tools/convert.py", line 36, in <lambda>
    safe_eval = lambda expr, ctx={}: s_eval(expr, ctx, nocopy=True)
  File "/.repo_requirements/odoo/openerp/tools/safe_eval.py", line 296, in safe_eval
    c = test_expr(expr, _SAFE_OPCODES, mode=mode)
  File "/.repo_requirements/odoo/openerp/tools/safe_eval.py", line 152, in test_expr
    expr = expr.strip()
ValueError: "'bool' object has no attribute 'strip'" while compiling
False
```

Maybe we should give a message why we cannot generate this planning

rafaelbn:
Hi @Yajo ,
In other use case like this one we also need a message about what's wrong:



Thanks

yajo:
https://github.com/OCA/crm/pull/174#issuecomment-331874363 happened when domain was empty, it should be fixed now.

https://github.com/OCA/crm/pull/174#issuecomment-331875472 happens because the incompatible `crm_phonecall_summary_predefined` addon is installed. This addon changes dramatically the way `crm.phonecall` object behaves without providing a proper compatibility, so it's not this PR's problem. I added a note in known issues. Remove it and test please. 😉

yajo:
This should go 100% coverage now.

chienandalu (approved):
Seems to work ok in runbot.

(I had to uninstall crm_phonecall_summary_predefined and the update this module to refresh the phonecall model)

crm_phonecall_planner/wizards/crm_phonecall_planner.py
pedrobaeza: I think this guideline is deprecated. Have you tried to inherit now in v9/v10 directly the method to see if it works?

yajo: Well, I remember to have tried in the past with a core model that defined a lambda, and I had to redefine the field just to override the default method.

The problem when you do `default=_default_duration` is that you bind directly the function code object. By using a lambda, you bind directly the lambda code, but inside it you have the benefits of all the inheritance system. That's the reason behind that guideline.

Other field attributes, such as `compute` or `search`, accept a string that defines the method name and make this wrapper automatically, but in `default` that system is not available, since passing a string would use that string as the default value instead, so the lambda callback wrapper is the best choice here AFAIK.

yajo:
@pedrobaeza @rafaelbn please update your reviews

crm_phonecall_planner/wizards/crm_phonecall_planner.py
pedrobaeza: OK, yes, it makes sense

pedrobaeza (approved):

crm_phonecall_planner/README.rst
```diff
@@ -0,0 +1,121 @@
...
+
+If you choose not to repeat calls, the planner will try to schedule one single
+phone call for each **criteria combination** (*Partner + Campaign + Source +
+Medium*) under the specified *Times* conditions.
+
+If you choose instead to repeat calls after some amount of days, the planner
+will try to schedule before all the partners that you never called under
+current **criteria combination**, but if none is found it will repeat calls for
+such **criteria combination** if last call was done before the minimum
+*Days gap* you specified.
```
rafaelbn: I don't undestand this paragraph 😢

rafaelbn:
Hi @Yajo:

Imagine you have a Phone call Schedule for:

- Tuesday from 09:00 to 14:00
- Thursday from 09:00 to 14:00

This are all the posibilities for Octobre 2017

- [ ] 10/03/2017 09:00:00	
- [ ] 10/05/2017 09:00:00
- [ ] 10/10/2017 09:00:00	
- [ ] 10/12/2017 09:00:00		
- [ ] 10/17/2017 09:00:00	
- [ ] 10/19/2017 09:00:00
- [ ] 10/24/2017 09:00:00
- [ ] 10/26/2017 09:00:00	
- [ ] 10/31/2017 08:00:00

And you would like you would like to plan call for every tuesday in october.

- [ ] How to do it? (for this opcion I expect only to lunch plannner/wizard one (1) time maximun)

- [ ] Day gap refers to the schedule o to the Worktime of the company?

- [ ] Which is the worktime of the company? Some companies work from monday to friday, some from monday to satudady morning, some from monday to saturday, some every day, some from sunday to thursday, some from friday to sunday, etc... So please could you add in README, which schedule take in care day gap?

- [ ] If you have a Schedule every day from monday to friday from 9:00 to 17:00. How to schedule for October every tuesday and thursday? (for this opcion I expect only to lunch plannner/wizar two times maximun).

Thanks

chienandalu:
@Yajo Maybe to achieve what @rafaelbn asks, the possibility of choosing an attending schedule that'd crop against the availability partner schedules would fit here and give way more flexibility.

yajo:
Day gap is only the amount of calendar days you are going to leave your target callee with no calls in between.

The easiest way to fix all your concerns would be just adding 7 checkboxes (one per week day) and let the user choose what week days to include in the planning, since the working daytime is chosen by the user directly in the wizard (no need to consult company or user working schedule).

However my concern is if this is a real use case... To begin, it makes no sense to record that your customer expects to be contacted on Sundays if you're never going to call him on Sundays, don't you think?

yajo:
So can this be merged as is and leave the rest of improvements for future iterations?

rafaelbn (changes requested):

yajo:
Better now @rafaelbn ?

rafaelbn (approved):
:+1: ready thank!

PR #170, opened 2017-09-05
[MIG] crm_claim: Migration to 10.0

cubells:
As crm_claim is not available in odoo v10.0, we migrate crm_claim to oca.

cc @Tecnativa

PR #177, opened 2017-09-30
[MIG] crm_claim: Migration to 10.0

bodedra:


pedrobaeza:
There's already a previous migration of that module in #170. Please review that one instead.

PR #178, opened 2017-10-01
[10.0][ENH] crm_claim

bodedra:
Would you please review following points:

1.    remove _rec_name = 'name' from crm.claim.stage object.

1.    Visible Teams in stage form view which will help to setup default team.

1.    Add OCA logo.


chienandalu (commented):
Why all the code style change?

pedrobaeza:
And not only that, you should make a PR to our branch, not a full new PR. Closing this.

pedrobaeza:
Sorry, the module was merged and it was incorrectly closed. I reopen, but please, don't make any stylistic change (that it's also against OCA guidelines).

crm_claim/README.rst
```diff
@@ -53,6 +53,7 @@ Contributors
 
 * Odoo S.A.
 * Vicent Cubells <vicent.cubells@tecnativa.com>
+* Bhavesh Odedra <bodedra@ursainfosystems.com>
```
mreficent: Don't put yourself as contributor because your contribution is purely stylistic and doesn't add/fix useful content.

crm_claim/__init__.py
```diff
@@ -2,5 +2,7 @@
 # Copyright 2015-2017 Odoo S.A.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-from . import models
-from . import report
+from . import (
+    models,
+    report,
+)
```
MiquelRForgeFlow: Please, delete this change, because I have some scripts that runs over `__init__` files and I don't want to change them just because you want to use this kind of notation that nobody uses.

crm_claim/models/__init__.py
```diff
@@ -2,7 +2,9 @@
...
-from . import crm_claim
-from . import crm_claim_category
-from . import crm_claim_stage
-from . import res_partner
+from . import (
+	crm_claim,
+	crm_claim_category,
+	crm_claim_stage,
+	res_partner,
+)
```
MiquelRForgeFlow: delete change

crm_claim/report/__init__.py
```diff
@@ -2,4 +2,6 @@
 # Copyright 2015-2017 Odoo S.A.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-from . import crm_claim_report
+from . import (
+    crm_claim_report,
+)
```
MiquelRForgeFlow: delete change

crm_claim/tests/__init__.py
```diff
@@ -2,4 +2,6 @@
 # Copyright 2017 Tecnativa - Vicent Cubells
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from . import test_crm_claim
+from . import (
+	test_crm_claim,
+)
```
MiquelRForgeFlow: delete change

MiquelRForgeFlow (changes requested):

crm_claim/__init__.py
yajo: Although I agree on not changing this to reduce diff, not sure what scripts do you have but those are wrong indeed, this is perfectly valid Python.

yajo (changes requested):
Please as said by others, we should try to keep changes to minimal.

If anything violates a linter rule, then change it, but otherwise it's better to leave it as it was, so we can review directly the important change (BTW, I can't distinguish if your pull request is actually adding something between all the cosmetic changes).

Besides, if you add a description to the PR, it also becomes easier to know what are we reviewers supposed to review.

Thanks!

bodedra:
I have revert back my changes.

Would you please review following points:

1.  remove `_rec_name = 'name'` from _crm.claim.stage_ object.

1.  Visible `Teams` in stage form view which will help to setup default team.

1.    Add OCA logo.

MiquelRForgeFlow (approved):

yajo (approved):
Yikes, still broken runbot. Code seems OK now (note: squash at merge)

PR #185, opened 2017-11-07
[10.0]crm_phonecall: Fix action to redirect to the new phonecall when its created from phonecall2phonecall wizard.

jesusVMayor:
Also improves the schedule_another_phonecall function to return a recordset with the new phonecalls instead of a dict of it.

### DIFF
--- a/crm_phonecall/models/crm_phonecall.py
+++ b/crm_phonecall/models/crm_phonecall.py
@@ -136,7 +136,8 @@ class CrmPhonecall(models.Model):
     @api.multi
     def schedule_another_phonecall(self, schedule_time, call_summary,
                                    user_id=False, team_id=False,
-                                   tag_ids=False, action='schedule'):
+                                   tag_ids=False, action='schedule',
+                                   return_recordset=False):
         """
         action :('schedule','Schedule a call'), ('log','Log a call')
         """
@@ -169,7 +170,10 @@ class CrmPhonecall(models.Model):
             if action == 'log':
                 call.write({'state': 'done'})
             phonecall_dict[call.id] = new_id
-        return phonecall_dict
+        if return_recordset:
+            return reduce(lambda x, y: x + y,  phonecall_dict.values())
+        else:
+            return phonecall_dict
 
     @api.onchange('opportunity_id')
     def on_change_opportunity(self):
--- a/crm_phonecall/wizard/crm_phonecall_to_phonecall.py
+++ b/crm_phonecall/wizard/crm_phonecall_to_phonecall.py
@@ -55,17 +55,16 @@ class CrmPhonecall2phonecall(models.TransientModel):
     def action_schedule(self):
         phonecall_obj = self.env['crm.phonecall']
         phonecalls = phonecall_obj.browse(self._context.get('active_ids', []))
-        for this in self:
-            phonecalls.schedule_another_phonecall(
-                this.date,
-                this.name,
-                this.user_id.id,
-                this.team_id.id or False,
-                this.tag_ids.ids,
-                action=this.action,
-            )
-
-        return phonecalls.redirect_phonecall_view()
+        new_phonecalls = phonecalls.schedule_another_phonecall(
+            self.date,
+            self.name,
+            self.user_id.id,
+            self.team_id.id or False,
+            self.tag_ids.ids,
+            action=self.action,
+            return_recordset=True
+        )
+        return new_phonecalls.redirect_phonecall_view()
 
     @api.model
     def default_get(self, fields):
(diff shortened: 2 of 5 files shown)


pedrobaeza:
Please change module version number. You are also changing API returning something different, which should be avoided in an already published module. Isn't possible to do it with "old" method? Please put a TODO for changing it for next version instead.

jesusVMayor:
fixed

crm_phonecall/wizard/crm_phonecall_to_phonecall.py
```diff
@@ -56,16 +56,18 @@ def action_schedule(self):
...
                 this.name,
                 this.user_id.id,
                 this.team_id.id or False,
                 this.tag_ids.ids,
                 action=this.action,
             )
-
-        return phonecalls.redirect_phonecall_view()
+        new_phonecalls = self.env['crm.phonecall']
+        for new_phonecall in new_phonecalls_dict.values():
```
pedrobaeza: You can do `new_phonecall = reduce(lambda x, y: a + b, new_phonecalls_dict.values())` for reducing code

jesusVMayor: done

pedrobaeza (approved):
@Yajo @chienandalu please review

crm_phonecall/wizard/crm_phonecall_to_phonecall.py
```diff
@@ -56,16 +56,17 @@ def action_schedule(self):
...
                 this.user_id.id,
                 this.team_id.id or False,
                 this.tag_ids.ids,
                 action=this.action,
             )
-
-        return phonecalls.redirect_phonecall_view()
+        new_phonecalls = reduce(lambda x, y: x + y,
+                                new_phonecalls_dict.values())
+        return new_phonecalls.redirect_phonecall_view()
```
yajo: If I'm not wrong, `new_phonecalls` is now a `list`, which shouldn't have a `redirect_phonecall_view()` method.

crm_phonecall/models/crm_phonecall.py
```diff
@@ -139,6 +139,7 @@ def schedule_another_phonecall(self, schedule_time, call_summary,
                                    tag_ids=False, action='schedule'):
         """
         action :('schedule','Schedule a call'), ('log','Log a call')
+        TODO: return a recordset instead of a dict
```
yajo: I think the fix is much easier if you just do this. Simply add a param `return_recordset=False`, and return a recordset if it is `True`, to keep backwards compatibility. Then add to Roadmap a note about removing that feature and always returning a recordset, as a good migration reminder.

yajo (changes requested):

crm_phonecall/wizard/crm_phonecall_to_phonecall.py
pedrobaeza: No, you are wrong :stuck_out_tongue_winking_eye: 

It returns a recordset. The reduce is for that.

yajo: How does that happen? 👀

pedrobaeza: Well, check reduce documentation

yajo: OK this code was not quite clear to me, but now I understand. It wasn't the problem with `reduce` (I know how that works bro 😆), it's more about the actual implementation... well, read above comment.

In any case, I'd have made just `sum(new_phonecalls_dict.values())`

crm_phonecall/wizard/crm_phonecall_to_phonecall.py
```diff
@@ -56,16 +56,17 @@ def action_schedule(self):
         phonecall_obj = self.env['crm.phonecall']
         phonecalls = phonecall_obj.browse(self._context.get('active_ids', []))
         for this in self:
-            phonecalls.schedule_another_phonecall(
+            new_phonecalls_dict = phonecalls.schedule_another_phonecall(
```
yajo: If I'm not wrong, you are here generating phone calls in an exponential fashion:

1. `phonecalls` contains N recordsets.
2. You iterate over `phonecalls` and call `schedule_another_phonecall()` on each of those with the current values, thus creating N * N phonecalls.
2. You only store the last call's result into `new_phonecalls_dict`, making it seem like only N phonecalls were created.

So either:

- Call `this.schedule_another_phonecall`
- Call only once (without `for` loop) `phonecalls.schedule_another_phonecall`
- I'm completely wrong and I don't understand what you're doing here

yajo (changes requested):

crm_phonecall/wizard/crm_phonecall_to_phonecall.py
pedrobaeza: `sum` works with recordsets? TIL then

jesusVMayor:
@Yajo fixed. I don't use the sum function, because it seems that it don't work with a list of recordset.

yajo (approved):

PR #195, opened 2018-03-21
[ADD][8.0] New module: email_template_configurator

andreparames:
This module implements a simpler interface for the email template dynamic fields. A configurator allows the creation of predefined named replacement strings, which can then be selected in the template.


andreparames:
Is this the correct repository, or should I submit it to OCA/social instead?

pedrobaeza:
I would say OCA/social

andreparames:
Thanks @pedrobaeza. I've opened a new PR: https://github.com/OCA/social/pull/247.

PR #196, opened 2018-03-21
[ADD] crm_industry: Establish link on crm lead to res.partner.industry

ghost:
With odoo v11 a new model has been introduced: res.partner.industry https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/res/res_partner.py#L818

this module adds the link to crm.lead similar to crm_sector module but stays within the standard of odoo since crm_sector depend on partner_sector (another oca module)

however, the model res.partner.sector is quite different since it has hierarchy and res.partner.industry has not

### DIFF
--- a/crm_industry/README.rst
+++ b/crm_industry/README.rst
@@ -1,0 +1,54 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+    :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+
+============
+CRM Industry
+============
+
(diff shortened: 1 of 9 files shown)


pedrobaeza:
Why don't you use crm_sector?

ghost:
to stay closer to odoo standard. it's a new model introduced with 11.0

pedrobaeza:
OK, thanks for the information. It makes sense.

ghost:
ok. :+1:  thanks for the input, next time will explain in the description :)

gdgellatly:
Closing, user has gone, module already merged

PR #204, opened 2018-06-08
[10.0] Create module crm_lead_partner_role

grindtildeath:

PR #205, opened 2018-06-08
[10.0] Create module crm_lead_project

grindtildeath:


crm_lead_project/models/crm_lead.py
```diff
@@ -0,0 +1,38 @@
...
+    def action_open_project_form(self):
+        return {
+            'name': _('Project'),
+            'type': 'ir.actions.act_window',
+            'res_model': 'project.project',
+            'view_type': 'form',
+            'view_mode': 'form',
+            'view_id': self.env.ref('project.edit_project').id,
+            'res_id': self.project_id.id,
+            'context': "{'remove_create_button': True}"
```
simahawk: use dict not string

crm_lead_project/wizard/crm_lead_to_opportunity.py
```diff
@@ -0,0 +1,51 @@
...
+        if self.partner_id:
+            values['partner_id'] = self.partner_id.id
+        leads = self.env['crm.lead'].browse(
+            self._context.get('active_ids', []))
+        # Convert to opportunity as it's done by odoo
+        values.update({'lead_ids': leads.ids, 'user_ids': [self.user_id.id]})
+        self._convert_opportunity(values)
+        # Create the project
+        project_values = self._prepare_project_values(leads)
+        project = self.env['project.project'].create(project_values)
```
simahawk: have you tried this as CRM user? Might need a couple of permissions update.
You can also check at the beginning of the method if the user is allowed to create projects and show a user error if not.

crm_lead_project/wizard/crm_lead_to_opportunity.py
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+    @api.multi
+    def action_apply(self):
+        self.ensure_one()
+
+        if self.name != 'convert_create_project':
+            return super(Lead2OpportunityPartner, self).action_apply()
+        # TODO Check if really needed
+        if len(self.env.context.get('active_ids')) > 1:
```
simahawk: can't we loop on them?

crm_lead_project/models/crm_lead.py
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+    @api.model
+    def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
```
simahawk: why don't you simply define a project view w/ no "create"?

simahawk (changes requested):

crm_lead_project/wizard/crm_lead_to_opportunity.py
grindtildeath: Shouldn't we instead simply create the `project.project` with sudo() in case the user has CRM access only ?
So he could still use the wizard to transform the lead into opportunity with the creation of project, but won't be able to interact with the project part.

crm_lead_project/models/crm_lead.py
grindtildeath: Because we would then need glue modules to display the fields brought in by hr.timesheet or project.forecast on the project form, if those are installed.
This way, we ensure the opened form view contains all the things from its inheritances.

crm_lead_project/wizard/crm_lead_to_opportunity.py
grindtildeath: Fixed in 031f2f6a5ce53aa903da339ca08f39148fa6d7f7

crm_lead_project/wizard/crm_lead_to_opportunity.py
```diff
@@ -45,7 +42,26 @@ def action_apply(self):
         values.update({'lead_ids': leads.ids, 'user_ids': [self.user_id.id]})
         self._convert_opportunity(values)
         # Create the project
-        project_values = self._prepare_project_values(leads)
-        project = self.env['project.project'].create(project_values)
-        leads.write({'project_id': project.id})
+        for lead in leads:
+            project_values = self._prepare_project_values(leads)
+            if self.env.user.has_group(
```
simahawk: @grindtildeath pls, check for permission instead otherwise even if you have rights you cannot add a project w/ your user.
NOTE: generally speaking checking for groups instead of permissions is not a good pattern.

crm_lead_project/wizard/crm_lead_to_opportunity.py
```diff
@@ -45,7 +42,26 @@ def action_apply(self):
...
+
+
+class Lead2OpportunityMassConvert(models.TransientModel):
+
+    _inherit = 'crm.lead2opportunity.partner.mass'
+
+    @api.onchange('name')
+    def _onchange_name(self):
+        """Restrict deduplication if a project is to create."""
+        if self.name == 'convert_create_project':
```
simahawk: self.deduplicate = self.name != 'convert_create_project'

simahawk:
@grindtildeath check builds (I guess they are failing for the same reasons as #204

crm_lead_project/models/project.py
```diff
@@ -37,3 +37,10 @@ def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
             root.set('create', 'false')
             res['arch'] = etree.tostring(root)
         return res
+
+
+class ProjectTask(models.Model):
+
+    _inherit = 'project.task'
+
+    crm_project = fields.Boolean(related='project_id.crm_project')
```
simahawk: readonly

simahawk: stored if you use it for searches but in this case you can use `project_id.crm_project` in domain and skip this new field

crm_lead_project/views/crm_lead.xml
```diff
@@ -12,6 +12,10 @@
                         icon="fa-puzzle-piece"
                         attrs="{'invisible': [('project_id', '=', False)]}" />
             </xpath>
+            <xpath expr="//header" position="inside">
+                <button name="%(crm_lead_project.crm_opportunity_create_project_action)d" string="Create project" type="action" class="btn-link"
```
simahawk: put `string="Create project" type="action" class="btn-link"` on a new line

crm_lead_project/wizard/crm_opportunity_create_project.py
```diff
@@ -0,0 +1,18 @@
...
+class CrmOpportunityCreateProject(models.TransientModel):
+
+    _inherit = 'crm.lead2opportunity.partner'
+    _name = 'crm.opportunity.create.project'
+
+    @api.multi
+    def action_create_project(self):
+        self.ensure_one()
+        leads = self.env['crm.lead'].browse(
+            self._context.get('active_ids', []))
```
simahawk: self.env.context

simahawk (changes requested):

rafaelbn:
Hi @grindtildeath @simahawk this PR is quite old, but if @simahawk make the review of his comments attended and all bots are green we can go ahead. If not please close the PR. Thanks!

simahawk (approved):

PR #207, opened 2018-06-13
[10.0] fix travis

grindtildeath:


crm_claim/report/crm_claim_report.py
```diff
@@ -159,12 +159,13 @@ def init(self):
...
             CREATE OR REPLACE VIEW %s AS (
                 %s
                 from
                 %s
                 %s
             )""" % (
             self._table, self._select(), self._from(), self._group_by(),
-        ))
+        )
+        self.env.cr.execute(sql)
```
yajo: I see you're trying to trick the linter.

Actually you can use this instead, which would be the correct way:


    from psycopg2.extensions import AsIs
    self.env.cr.execute("""
        CREATE OR REPLACE VIEW %s AS (
                 %s
                 from
                 %s
                 %s
             )""", (
             AsIs(self._table), AsIs(self._select()), AsIs(self._from()), AsIs(self._group_by()),
    ))

crm_phonecall/report/crm_phonecall_report.py
```diff
@@ -125,8 +125,9 @@ def _from(self):
     def init(self):
 
         tools.drop_view_if_exists(self._cr, self._table)
-        self._cr.execute("""
+        sql = """
             create or replace view %s as (
                 %s
                 %s
-            )""" % (self._table, self._select(), self._from()))
+            )""" % (self._table, self._select(), self._from())
```
yajo: Use `AsIs`

yajo (changes requested):

crm_claim/report/crm_claim_report.py
grindtildeath: Thank you @Yajo  :)

grindtildeath:
Rebased 
@Yajo can you update your review ? 
@pedrobaeza Can we fast-track ? These errors are making travis fail on 10.0 PRs (Cf #204, #205 )

pedrobaeza (approved):

PR #208, opened 2018-06-20
[9.0] add crm_meeting_commercial_partner

JordiBForgeFlow:
CRM Meeting Commercial Partner
==============================

This module allows to display the meetings of a company as the sum of all
the meetings held with the company's contacts.

Usage
=====
Go to the partner views. You will see, in the Kanban and Form views, that
the number of meetings shown for a company includes the meetings held with
the company's contacts.

PR #209, opened 2018-06-20
[10.0][mig][crm_meeting_commercial_partner]

JordiBForgeFlow:
Migration from https://github.com/OCA/crm/pull/208

CRM Meeting Commercial Partner
==============================

This module allows to display the meetings of a company as the sum of all
the meetings held with the company's contacts.

Usage
=====
Go to the partner views. You will see, in the Kanban and Form views, that
the number of meetings shown for a company includes the meetings held with
the company's contacts.


pedrobaeza:
What is the advantage of this over checking the "All" calendar?

JordiBForgeFlow:
@pedrobaeza When you go to a partner, say "Eficent", you want to see how many meetings are scheduled with Eficent's individual persons, because as a general rule no one meets with Eficent as a company.

pedrobaeza:
OK, got it. I was thinking on the company, not on the partner side. But this is not correctly put nor named. It should be `calendar_commercial_partner` or `calendar_commercial_partner_meeting`, and be relocated to OCA/calendar, not here.

JordiBForgeFlow:
But this feature depends on module `crm`, not `calendar.` That's why I suggested it here

pedrobaeza:
I think you should go more for the generic one (these CRM meetings are also calendar meetings).

JordiBForgeFlow:
@pedrobaeza This module builds on top of a feature to display the partner's meetings that is introduced by odoo in the CRM module.
https://github.com/odoo/odoo/blob/10.0/addons/crm/models/res_partner.py#L37
https://github.com/odoo/odoo/blob/10.0/addons/crm/views/res_partner_views.xml#L119

It's all over the crm module.

What this module does is to extend the methods to allow the user to view all the meetings from a company perspective, to have an overall view of all meetings scheduled with individual contacts.

By removing the dependency with crm we would duplicate the features.

pedrobaeza:
Ok, let's go this way.

LoisRForgeFlow (approved):
LGTM :+1:

JordiBForgeFlow:
@bjeficent @ageficent can you test this on runbot? @lreficent can help you.

crm_partner_schedule_meeting/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# -*- encoding: utf-8 -*-
```
AdriaGForgeFlow: I think this file does not belong here. It makes Odoo crash. Solved deleting the folder.

AdriaGForgeFlow (changes requested):
Odoo crashing on startup due to crm_partner_shcedule_meeting/ folder.

LoisRForgeFlow:
@ageficent It's fixed, could you update your review?

AdriaGForgeFlow (approved):
LGTM

PR #214, opened 2018-07-17
[11.0][IMP] crm_claim: add activities to claims

oihane:


anajuaristi (approved):

crm_claim/migrations/11.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    activity_type = env.ref('mail.mail_activity_data_todo')
+    crm_claim_model = env.ref('crm_claim.model_crm_claim')
+    cr.execute("""
+        SELECT
```
pedrobaeza: You can make the insert from a select query and make this faster

pedrobaeza (changes requested):

crm_claim/migrations/11.0.2.0.0/pre-migration.py
oihane: Change done

crm_claim/migrations/11.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,33 @@
...
+          user_id,
+          date_action_next,
+          action_next,
+          %s
+        FROM
+          crm_claim
+        WHERE
+          date_action_next IS NOT Null
+        OR
+          action_next IS NOT Null;
```
pedrobaeza, line 32: Is it possible to have activities without due date?

oihane: No, you're right didn't think about it

oihane: @pedrobaeza fixed

crm_claim/migrations/11.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,34 @@
...
+         mail_activity
+           (res_model, res_model_id, res_id, res_name, user_id,
+            date_deadline, summary, activity_type_id)
+        SELECT
+          %s,
+          %s,
+          id,
+          name,
+          user_id,
+          CASE WHEN date_action_next IS NOT Null THEN date_action_next
```
pedrobaeza: Use `COALESCE(date_action_next, now())` instead for simplifying the query

oihane: Done

crm_claim/migrations/11.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,33 @@
...
+          id,
+          name,
+          user_id,
+          COALESCE(date_action_next, now()),
+          action_next,
+          %s
+        FROM
+          crm_claim
+        WHERE
+          date_action_next IS NOT Null
```
pedrobaeza, line 30: I'm thinking that you can remove now this part of the WHERE

oihane: Before this, you can set a 'date_action_next' without adding an 'action_next', if you don't add this where it won't create any activity (a 'summary' is not mandatory in the database, you can create one without text)

pedrobaeza: Is the text mandatory on activities?

oihane: There are 3 fields mandatory through odoo:
* user_id
* date_deadline
* activity_type_id

Internally:
* res_model
* res_model_id
* res_id
* res_name

`summary` and `note` are NOT MANDATORY

pedrobaeza: OK, thanks for the confirmation.

pedrobaeza (approved):

PR #220, opened 2018-08-22
[IMP] crm_claim: autogenerated readme

oihane:


pedrobaeza (approved):
Fast-tracking as this is a README by fragments ref

PR #224, opened 2018-09-10
[11.0][MIG] calendar_resource

eilst:
Migration of calendar_resource to 11.0


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

pedrobaeza:
I think this should be in OCA/calendar instead of here.

PR #213, opened 2018-07-12
[MIG] crm_phonecall_planner: Migration to 11.0

chienandalu:
* Minor migration tweaks.
* Split into OCA readme parts

Depends on:

- [x] crm_phonecall: https://github.com/OCA/crm/pull/212
- [x] partner_phonecall_schedule: https://github.com/OCA/partner-contact/pull/606

cc @Tecnativa

PR #228, opened 2018-09-18
[FIX] crm_phonecall_planner: fix domain widget

chienandalu:
Fixes the partner domain filter as reported here: https://github.com/OCA/crm/pull/213#issuecomment-422045941

cc @Tecnativa


pedrobaeza (approved):
Fast-tracking as obvious

PR #234, opened 2018-10-18
[11.0][MIG] marketing_crm_partner: Migration to 11.0

cristinamartinrod:
This module extends the functionality of the CRM to support having the tracking fields available in the partner and copy them there automatically when the partner is created from a lead/opportunity.

cc @Tecnativa


cristinamartinrod:
**Fix Test - WIP**

.gitignore
```diff
@@ -54,3 +54,4 @@ docs/_build/
 # Backup files
 *~
 *.swp
+marketing_crm_partner/.DS_Store
```
simahawk: this should be *.DS_Store

simahawk (approved):
LG. minor issue w/ .gitignore

simahawk:
tests do not pass.

simahawk:
@cristinamartinrod pls add [WIP] to title :)

rafaelbn (approved):

marketing_crm_partner/models/crm_lead.py
```diff
@@ -0,0 +1,21 @@
...
+class CRMLead(models.Model):
+    _inherit = "crm.lead"
+
+    @api.model
+    def _create_lead_partner_data(self, name, is_company, parent_id=False):
+        """Populate marketing fields in partner."""
+        res = super()._create_lead_partner_data(
+            name, is_company, parent_id=parent_id,
+        )
+        for _key, field, _cookie in self.tracking_fields():
```
simahawk: @pedrobaeza on my 1st review I was about to ask for this change (use self.tracking_fields) but then I read the code in utm.mixin where it's clearly stated that we are not supposed to use the one on self but the one on the mixin straight (as it was before). 

Check [this](https://github.com/odoo/odoo/blob/11.0/addons/utm/models/utm.py#L47).
If you have a better insight... I'd love to hear that! :wink:

.gitignore
cristinamartinrod: Done!

marketing_crm_partner/models/crm_lead.py
pedrobaeza: You are right, Simone! I switch back to `self.env['utm.mixin'].tracking_fields()` for not preventing overridings over the mixin.

pedrobaeza:
Merging this

PR #251, opened 2018-12-19
[11.0][IMP]meeting count calculation performance

AaronHForgeFlow:
Improved performance almost 4x faster.

### DIFF
--- a/crm_meeting_commercial_partner/models/res_partner.py
+++ b/crm_meeting_commercial_partner/models/res_partner.py
@@ -8,21 +8,25 @@ from odoo import api, fields, models
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
-    meeting_count = fields.Integer(compute='_compute_meeting_count')
+    meeting_count = fields.Integer(
+        compute='_compute_meeting_count_commercial_partner')
 
     @api.multi
-    def _compute_meeting_count(self):
-        super()._compute_meeting_count()
+    def _compute_meeting_count_commercial_partner(self):
+        att_model = self.env['calendar.attendee']
         for partner in self:
-            partners = partner + partner.child_ids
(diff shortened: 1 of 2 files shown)


JordiBForgeFlow:
@aheficent I have introduced a few changes. Still, I am not sure if this will introduce a performance boost. We'll need to evaluate.

AaronHForgeFlow:
@jbeficent I noticed of a very better performance when accessing contacts in a very big database. About 4x faster.

Edit: Hold on, I have not updated your changes.

AaronHForgeFlow:
Confirmed the performance improvement for this single computation method.

JordiBForgeFlow (approved):

LoisRForgeFlow (approved):
LGTM

crm_meeting_commercial_partner/models/res_partner.py
```diff
@@ -8,21 +8,22 @@
...
-    def _compute_meeting_count(self):
-        super()._compute_meeting_count()
+    def _compute_meeting_count_commercial_partner(self):
+        event_model = self.env['calendar.event']
         for partner in self:
-            partners = partner + partner.child_ids
-            meetings = self.env['calendar.event'].search(
-                [('partner_ids', 'in', partners.ids)])
-            partner.meeting_count = len(meetings)
+            partner.meeting_count = event_model.search_count(
```
pedrobaeza: Use `read_group` instead for total performance boost. @mreficent has done it before in for example: https://github.com/OCA/contract/commit/402370f1cbe5cde6b484200f87596e9d803a36af#diff-3dc0d52ea9449a93459ef5b7fc36a060R22

AaronHForgeFlow: Thanks @pedrobaeza ! It is much faster now! :)

PR #253, opened 2019-01-15
[12.0] [MIG] crm lead code

keshrath:


keshrath:
@oca-clabot I have sent an email to cla@odoo-community.org

crm_lead_code/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+    cr.execute('UPDATE crm_lead '
+               'SET code = id;')
+
+
+def assign_old_sequences(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, dict())
+    lead_obj = env['crm.lead']
+    sequence_obj = env['ir.sequence']
+    leads = lead_obj.search([], order="id")
+    for lead_id in leads.ids:
```
sbejaoui, line 21: Why using sql query Here?

```python
    for lead in leads:
         lead.code = sequence_obj.next_by_code('crm.lead')
```

crm_lead_code/models/crm_lead.py
```diff
@@ -0,0 +1,31 @@
...
+##############################################################################
+
+from odoo import api, fields, models, _
+
+
+class CrmLead(models.Model):
+    _inherit = "crm.lead"
+
+    code = fields.Char(
+        string='Lead Number', required=True, default="/", readonly=True)
```
sbejaoui, line 12: If it is readonly, why it is required? IMO, required is not necessary and we can remove the pre_init_hook.

sbejaoui (changes requested):
Thank you for this PR, 

can you check my comments please, Thanks

crm_lead_code/models/crm_lead.py
keshrath: The readonly attributes only ensures that this field is automatically displayed as readonly in any view.  It is also possible to create a lead via Python code or RPC.  In this case readonly has no meaning but the required flag ensures that null can't be passed to the create method.

crm_lead_code/__init__.py
keshrath: Since I only migrated the module and didn't rewrite it, I tried to change as little as possible to make future bugfixes etc easier in all versions.  In my opinion I would only change things if they bring a clear advantage or are necessary due to Odoo version changes.  Where do you see the advantage of not using SQL directly but using the ORM?  In this case performance doesn't matter as it is executed only once, but this way you make sure that there are no unwanted side effects through the ORM. Like setting the write_date and write_uid.

crm_lead_code/__manifest__.py
```diff
@@ -22,9 +21,10 @@
 
 {
     "name": "Sequential Code for Leads / Opportunities",
-    "version": "10.0.1.0.0",
+    "version": "12.0.1.0.0",
     "category": "Customer Relationship Management",
-    "author": "OdooMRP team, "
+    "author": "MuK IT GmbH, "
```
pedrobaeza: This migration doesn't grant you co-authorship. Please include yourself in contributors, not here.

crm_lead_code/__manifest__.py
```diff
@@ -22,9 +21,10 @@
...
+              "OdooMRP team, "
```
pedrobaeza: This is an old denomination. You can remove it.

crm_lead_code/__manifest__.py
```diff
@@ -22,9 +21,10 @@
...
               "AvanzOSC, "
               "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
```
pedrobaeza: Please switch to my new company, `Tecnativa`, instead

pedrobaeza (changes requested):

crm_lead_code/__init__.py
pedrobaeza: @sbejaoui this was done by SQL on purpose, as we want to avoid triggering any ORM code assigning this, as it's an initialization. This is also done for performance reasons.

sbejaoui: > Where do you see the advantage of not using SQL directly but using the ORM?

IMO, any use of sql queries should be justified otherwise the ORM is here to abstract the sql level and should be used in all cases by default.

I think that the sql query here can be justified by the need of avoiding `write_uid` update and there in not any computed field depends on this one.

crm_lead_code/models/crm_lead.py
sbejaoui: I was asking this because of the default value too. but you are right, as required here prevent null value to be forced.

sbejaoui (approved):
LGTM, thank you.

crm_lead_code/__manifest__.py
keshrath: https://github.com/OCA/crm/compare/3de96a73e5319c27372006aeca3c79db42804a91..95a25015a52ec101b840f55ac7e8d36dec32234c

pedrobaeza: Thanks! You can remove as well "Serv. Tecnol...", as it's the old company.

keshrath: https://github.com/OCA/crm/commit/81893f1c15023e12d96a9d5cb18876686540743e

keshrath: https://github.com/OCA/crm/compare/95a25015a52ec101b840f55ac7e8d36dec32234c..81893f1c15023e12d96a9d5cb18876686540743e

pedrobaeza (approved):

PR #254, opened 2019-01-23
8.0 mig mail thread track x2many

daramousk:


pedrobaeza:
This should be in OCA/social instead, and there's already a module proposed for this for v11: https://github.com/OCA/social/pull/254

hbrunn:
@pedrobaeza agreed with the repo, but this does something totally different than you think

hbrunn:
actually, this might better be a candidate for server-tools or server-backend

hbrunn:
@daramousk it helps when you paste the relevant parts of the readme in the first comment

PR #257, opened 2019-02-13
[10.0] [ADD] New module publication

NL66278:


pedrobaeza:
Should this be here or in some vertical?

NL66278:
@pedrobaeza I do not see any specific industry. We first developed this for a church organisation, and now improved and migrated it for a renters union. But I think it could be used by any ngo, government institution or company that has some regular publication to which you can subscribe.

pedrobaeza:
OK, difficult to classify... but CRM might not be the best one. Any other opinion here?

hbrunn:
I think CRM is good - this kind of stuff is meant to keep customer (member, etc) engagement, that's what CRM does.

publication/README.rst
```diff
@@ -0,0 +1,64 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+===========
+publication
```
elicoidal:
```suggestion
Publication
```

publication/README.rst
```diff
@@ -0,0 +1,64 @@
...
+===========
+
+This module allows to register publications that a partner can receive.
+
+Publications can be in print of electronic, and be free or paid for.
```
elicoidal:
```suggestion
Publications can be of multiple kind: printed, electronic, and free or paid for.
```

publication/README.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Paid publications can be linked to a contract (recurring analytic account).
+
+Usage
+=====
+
+To use this module, you need to:
```
elicoidal: adapt

publication/README.rst
```diff
@@ -0,0 +1,64 @@
...
+
+#. Go to ...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/{repo_id}/{branch}
```
elicoidal: adapt

publication/README.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Known issues / Roadmap
```
elicoidal: remove the section

publication/README.rst
```diff
@@ -0,0 +1,64 @@
...
+======================
+
+None at present.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
elicoidal: adapt

publication/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Copyright 2014-2018 Therp BV <https://therp.nl>.
```
elicoidal: 2019?

publication/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    'version': '10.0.1.0.0',
+    'category': 'Customer Relationship Management',
+    'author': 'Therp BV, '
+              'Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'summary': 'Maintain electronic or print publications for your relations.',
+    'depends': [
+        'partner_noncommercial',
+        'contract_line_extended',
+        'web_m2x_options',
```
elicoidal: do you need to adapt the oca_dependencies of the repo?

publication/i18n/nl.po
```diff
@@ -0,0 +1,387 @@
...
+#: model:ir.ui.view,arch_db:publication.view_partner_form
+msgid "Subscription contract lines"
+msgstr "Contractregels voor abonnementen"
+
+#. module: publication
+#: model:ir.ui.view,arch_db:publication.view_partner_form
+msgid "Publications to receive"
+msgstr "Te ontvangen publicaties"
+
+#~ msgid "Can be send by email"
```
elicoidal: remove dead code

publication/models/distribution_list.py
```diff
@@ -0,0 +1,180 @@
...
+ AND c.partner_id = %s"""
+
+SQL_ASSIGNED_COUNT = """
+SELECT COALESCE(SUM(copies), 0) as copies
+ FROM distribution_list
+ WHERE product_id = %s
+ AND contract_partner_id = %s"""
+
+
+class DistributionList(models.Model):
```
elicoidal, line 76: Why not `PublicationList`?
It is a bit confusing to use Distribution and Publication everywhere

publication/models/distribution_list.py
```diff
@@ -0,0 +1,180 @@
...
+        string='Date end')
+    copies = fields.Integer(string='Number of copies', default=1)
+    name = fields.Char(
+        compute='_compute_name_address',
+        string='Name')
+    contact_address = fields.Char(
+        compute='_compute_name_address',
+        string='Receiving address')
+    contract_count = fields.Integer(
+        string="Number to distribute",
```
elicoidal: Quantity to distribute
Quantity already Assigned
Quantity still Available

publication/models/distribution_list.py
```diff
@@ -0,0 +1,180 @@
...
+        """Sets the proper domain for contract_partner.
+
+        Also enforces first selecting the publication.
+        """
+        self.ensure_one()
+        if self.contract_partner_id and not self.partner_id:
+            self.partner_id = self.contract_partner_id
+        if self.contract_partner_id and not self.product_id:
+            raise ValidationError(_(
+                "You must select a publication, before selecting"
```
elicoidal:
```suggestion
                "You must select a publication before selecting"
```

publication/models/distribution_list.py
```diff
@@ -0,0 +1,180 @@
...
+        return result
+
+    @api.multi
+    def _limit_count(self):
+        """Limit number of copies send to amount set in contract lines.
+
+        It should be possible to make a constrains method of this function,
+        but for inexplicable reasons this does not work.
+        """
+        for this in self:
```
elicoidal: this -> copy ?

publication/models/product_template.py
```diff
@@ -0,0 +1,29 @@
...
+            ('daily', 'Day(s)'),
+            ('weekly', 'Week(s)'),
+            ('monthly', 'Month(s)'),
+            ('quarterly', 'Quarterly'),
+            ('yearly', 'Year(s)')],
+        default='monthly',
+        string='Publishing frequency',
+        help="At what intervals publication is published")
+    publishing_frequency_interval = fields.Integer(
+        string='Publish Every',
```
elicoidal:
```suggestion
        string='Published Every',
```

publication/models/product_template.py
```diff
@@ -0,0 +1,29 @@
...
+            ('irregular', 'Irregular'),
+            ('daily', 'Day(s)'),
+            ('weekly', 'Week(s)'),
+            ('monthly', 'Month(s)'),
+            ('quarterly', 'Quarterly'),
+            ('yearly', 'Year(s)')],
+        default='monthly',
+        string='Publishing frequency',
+        help="At what intervals publication is published")
+    publishing_frequency_interval = fields.Integer(
```
elicoidal, line 26: Is the day of reception important (eg: every Thursday or every day 1 of the month)?

publication/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_distribution_list_manager,CRUD access for distribution.list,model_distribution_list,base.group_user,1,1,1,1
```
elicoidal: Do we need to distinguish between user (1000)/manager (1111)?

elicoidal (commented):
Thanks for the contribution!
no test

publication/models/distribution_list.py
NL66278: It is DistributionList because it is not a list of publications, but rather a list about how publications should be distributed over recipients.

publication/models/distribution_list.py
NL66278: We always use this when iterating over self. In this case it would also be confusing as we are not iterating over copies but over distribution list records, that have copies as a field.

publication/models/product_template.py
NL66278: For the moment I do not see a use for this.

NL66278:
@elicoidal I followed up most of your suggestions, unless otherwise noted.

publication/models/distribution_list.py
hbrunn: I'd still propose to rename this to `publication.distribution.list` to avoid possible clashes with other model names that are distribution lists, as that's a quite generic concept. Personally, I usually prefix all model names an addon introduces with the addon's name, as those have to be unique anyways. Produces sometimes horribly long model names, but it unique.

publication/models/account_analytic_invoice_line.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+# Copyright 2014-2019 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+# pylint: disable=no-member,too-few-public-methods
```
elicoidal: To be removed?

elicoidal (approved):
LGTM

publication/models/account_analytic_invoice_line.py
NL66278: These are there to prevent spurious linting errors. pylint does not understand the Odoo inheritance mechanism, therefore does not see the members and methods inherited from other models.

NL66278:
I am considering renaming the module to contract_publication and moving it to the contract repository. To do this I will need to drop the dependency on partner_non_commercial. Then the module will only contain on other modules in the contract repository, and not on any other oca repo.

What do people think about this?

NL66278:
The alternative PR is here: https://github.com/OCA/contract/pull/277 I still would like people's opinion on this move to contract, to avoid increasing the dependencies between repo's, and because the contract model plays a central role in the proposed module.

claudiagn (approved):
i think you should move this to contract! lgtm! 👍

NL66278:
Closed in favour of this one: https://github.com/OCA/contract/pull/277

PR #264, opened 2019-04-12
[11.0][MIG] calendar_resource

eilst:
Migration with commit  history included. The PR https://github.com/OCA/crm/pull/224 that I did was not following some required steps for the migration specified in: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

https://github.com/OCA/crm/pull/224#issuecomment-421115900: > I think this should be in OCA/calendar instead of here.
Hey @pedrobaeza 
It seems reasonable to me also, but on branch 10.0 it was here. Should we move it?


pedrobaeza:
Yes, please move it!

PR #265, opened 2019-04-29
[ADD][12.0] crm_lead_currency

TDu:
This module allows to select a specific currency for an opportunity.
If the selected currency is different to the company currency,
an amount in the customer currency can be set and the expected revenue
of the opportunity will automatically be computed in the company currency.


crm_opportunity_currency/README.rst
```diff
@@ -0,0 +1,77 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
simahawk: pls use new readme structure for new modules

crm_opportunity_currency/README.rst
```diff
@@ -0,0 +1,77 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========================
+Crm Opportunity Currency
+========================
+
+This module allows to select a specific currency for an opportunity.
+If the selected currency is different to the company currency, an amount in the customer
```
simahawk: *different from

crm_opportunity_currency/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'CRM Opportunity Currency',
+    'summary': """
+        On opportunities add the amount in the company currency.""",
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'Camptocamp SA,Odoo Community Association (OCA)',
+    'website': 'www.camptocamp.com',
```
simahawk: https://github.com/OCA/crm

crm_opportunity_currency/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'depends': [
+        'crm',
+    ],
+    'data': [
+        'views/crm_lead_opportunity_currency_views.xml',
+    ],
+    'demo': [
```
simahawk: remove empty key

crm_opportunity_currency/models/crm_lead.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @api.multi
+    def get_revenue_in_company_currency(self):
+        """Compute the planned revenue in the company currency.
+
+        If the customer currency is different than the company currency,
+        the planned revenue is computed in the company currency.
+        """
+        self.ensure_one()
+        if self.company_currency == self.customer_currency_id:
```
simahawk: `if self.is_same_currency`?

crm_opportunity_currency/models/crm_lead.py
```diff
@@ -0,0 +1,66 @@
...
+        currency_field='customer_currency_id',
+        store=True,
+    )
+    is_same_currency = fields.Boolean(
+        string='Same currency',
+        compute='_compute_is_same_currency',
+    )
+
+    @api.onchange('customer_currency_id', 'amount_customer_currency')
+    def compute_expected_amount(self):
```
simahawk: `_onchange_currency`

crm_opportunity_currency/models/crm_lead.py
```diff
@@ -0,0 +1,66 @@
...
+            )
+
+    @api.multi
+    def write(self, vals):
+        res = super().write(vals)
+        if (
+            'customer_currency_id' in vals
+            or 'amount_customer_currency' in vals
+        ):
+            # Due to the readonly in the view the compute does not work
```
simahawk: I'd remove the readonly... I don't see any bad side effect... if they edit it, up to them to fix it, no?

simahawk (changes requested):

crm_opportunity_currency/README.rst
TDu: I think both are valid english

crm_opportunity_currency/__manifest__.py
TDu: Changed

crm_opportunity_currency/__manifest__.py
TDu: Done

crm_opportunity_currency/models/crm_lead.py
TDu: Yes

crm_opportunity_currency/models/crm_lead.py
TDu: Changed

crm_opportunity_currency/README.rst
TDu: Changed

crm_opportunity_currency/models/crm_lead.py
TDu: I would think that if it is not set to readonly the currency conversion should be handled in both direction.

i-vyshnevska: why don't relay planned_revenue on compute method

i-vyshnevska (commented):
would be good to send choosen currency from lead to SO, but it would call dependency from 'sale'

crm_opportunity_currency/models/crm_lead.py
```diff
@@ -0,0 +1,66 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import api, fields, models
+
+
+class CrmLead(models.Model):
+    _inherit = 'crm.lead'
+
+    customer_currency_id = fields.Many2one(
+        string='Customer currency',
```
luistorresm: In v12 when the string is the same that the field name, this parameter is not necessary.

And if it has `_id` or `_ids`, that is not added in the string.

Could you remove this parameter, please?

crm_opportunity_currency/models/crm_lead.py
```diff
@@ -0,0 +1,66 @@
...
+        store=True,
+    )
+    is_same_currency = fields.Boolean(
+        string='Same currency',
+        compute='_compute_is_same_currency',
+    )
+
+    @api.onchange('customer_currency_id', 'amount_customer_currency')
+    def _onchange_currency(self):
+        for lead in self:
```
luistorresm: In the onchange method is not necessary to make for, this is not multi.

[Here](https://www.odoo.com/documentation/10.0/reference/orm.html#odoo.api.onchange) the Odoo doc

crm_opportunity_currency/models/crm_lead.py
TDu: But it doesn't hurt to have it explicitly stated ?

crm_opportunity_currency/models/crm_lead.py
TDu: Thanks changed.

crm_opportunity_currency/models/crm_lead.py
```diff
@@ -0,0 +1,65 @@
...
+
+    customer_currency_id = fields.Many2one(
+        string='Customer currency',
+        comodel_name='res.currency',
+        default=lambda self: self.env.user.company_id.currency_id,
+    )
+    amount_customer_currency = fields.Monetary(
+        string='Customer amount',
+        currency_field='customer_currency_id',
+        store=True,
```
grindtildeath: I don't think this is necessary if the field is not computed.

crm_opportunity_currency/models/crm_lead.py
```diff
@@ -0,0 +1,65 @@
...
+        the planned revenue is computed in the company currency.
+        """
+        self.ensure_one()
+        if self.is_same_currency:
+            return self.planned_revenue
+        return self.customer_currency_id._convert(
+            self.amount_customer_currency or 0,
+            self.company_currency,
+            self.env.user.company_id,
+            fields.Datetime.now(),
```
grindtildeath: Would be great to somehow be able to change the date used here, but a simple mention in the roadmap section of the readme is fine at the moment.

crm_opportunity_currency/models/crm_lead.py
```diff
@@ -0,0 +1,65 @@
...
+    @api.multi
+    def write(self, vals):
+        res = super().write(vals)
+        if (
+            'customer_currency_id' in vals
+            or 'amount_customer_currency' in vals
+        ):
+            # Due to the readonly in the view the compute does not work
+            for lead in self:
+                lead.write({
```
grindtildeath: Calling a write inside a loop inside a write is usually a bad design.
I'd add `customer_currency_id=None` and `amount_customer_currency=None` to the params of `get_revenue_in_company_currency`. Then use `or` in the method to get it right, eg:

```
customer_currency = self.env['res.currency'].browse(customer_currency_id) or self.customer_currency_id
return customer_currency._convert(
    amount_customer_currency or self.amount_customer_currency or 0,
    (...)
```
Then move the loop before the call to super and update `vals` for each record, e.g.:
``` 
@api.multi
def write(self, vals):
    if ('customer_currency_id' in vals or 'amount_customer_currency' in vals):
        for lead in self:
            planned_revenue = lead.get_revenue_in_company_currency(
                customer_currency_id=vals.get('customer_currency_id')),
                amount_customer_currency=vals.get('amount_customer_currency')
            )
            vals['planned_revenue'] = planned_revenue
            super(CrmLead, lead).write(vals)
        return True
    else:
        return super().write(vals)
```

crm_opportunity_currency/views/crm_lead_opportunity_currency_views.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0"?>
+
+<odoo>
+
+  <record id="crm_case_form_view_oppor" model="ir.ui.view">
```
grindtildeath: Somehow these views should be restricted to group `base.group_multi_currency`

grindtildeath (changes requested):
Small changes

crm_opportunity_currency/models/crm_lead.py
TDu: Correct left over from previous code. Removed now

crm_opportunity_currency/models/crm_lead.py
TDu: Readme updated.

crm_opportunity_currency/views/crm_lead_opportunity_currency_views.xml
TDu: I am not sure it seems to me, that group is mainly used in accounting module to restrict some access. Here I think if the module is installed we want every user to have access to this feature.

crm_opportunity_currency/models/crm_lead.py
```diff
@@ -0,0 +1,64 @@
...
+from odoo import api, fields, models
+
+
+class CrmLead(models.Model):
+    _inherit = 'crm.lead'
+
+    customer_currency_id = fields.Many2one(
+        string='Customer currency',
+        comodel_name='res.currency',
+        default=lambda self: self.env.user.company_id.currency_id,
```
sebalix: Maybe add `ondelete='restrict`, what do you think? (the default value is `set null`)

crm_opportunity_currency/models/crm_lead.py
gurneyalex: it does not hurt at all

crm_opportunity_currency/models/crm_lead.py
gurneyalex: @sebalix yes (but in this specific case, chances are really high that some other required currency_id field on another model will prevent deleting the currency record, which is not something you do very often :smile_cat: )

crm_opportunity_currency/models/crm_lead.py
```diff
@@ -0,0 +1,64 @@
...
+            return self.planned_revenue
+        return self.customer_currency_id._convert(
+            self.amount_customer_currency or 0,
+            self.company_currency,
+            self.env.user.company_id,
+            fields.Datetime.now(),
+        )
+
+    @api.multi
+    def _compute_is_same_currency(self):
```
gurneyalex: needs changing: missing api.depends decorator -> the field is not recomputed when the currency is changed -> the field amount_customer_currency is not visible until you save the record

crm_opportunity_currency/models/crm_lead.py
```diff
@@ -0,0 +1,64 @@
...
+        for lead in self:
+            lead.is_same_currency = (
+                lead.customer_currency_id == lead.company_currency
```
gurneyalex: The test must be updated to return True if customer_currency_id is not set (in this case we assume it means we are using the company currency)

gurneyalex (changes requested):

pedrobaeza:
Following model naming, this should be `crm_lead_currency`

gurneyalex:
renaming requested by Pedro is in https://github.com/TDu/crm/pull/1 (with a small improvement in the manifest file)

crm_opportunity_currency/models/crm_lead.py
TDu: This has been done

crm_opportunity_currency/models/crm_lead.py
TDu: The change on the write has been removed following changes in the view to handle the problem with the readonly field.

crm_opportunity_currency/models/crm_lead.py
TDu: @gurneyalex I have made a slight variation see the last commit

TDu:
The module name has been changed I squashed the commit a little as well.

crm_opportunity_currency/models/crm_lead.py
TDu: It allows the module to works the same in creation or update mode.

gurneyalex (approved):

i-vyshnevska (approved):

claudiagn (approved):
LGTM!

PR #269, opened 2019-05-14
Fix computation of claim count when permissions are changed.

tbaptista:
If users don't have permission to read claims, they would be locked out of opening up the partner form view. This fix enables one to only allow claims to be viewed by e.g. Managers and not regular CRM users.


pedrobaeza:
If there's an access problem due to this field, the proper solution is not this one, but protecting whole view that is adding the field in the view with the minimum group that has access to the object.

tbaptista:
Thanks for the comment, I will look into doing it that way. I got the idea of the solution from the implementation of hr_contracts in the core Odoo.

PR #271, opened 2019-05-20
[MIG] crm_location: Migration to 12.0

Tardo:
Descriptive title


crm_location/tests/test_crm_location.py
```diff
@@ -18,18 +18,23 @@ def setUpClass(cls):
             'code': 'Test state code',
             'country_id': cls.country.id,
         })
-        cls.location = cls.env['res.better.zip'].create({
-            'name': '12345',
-            'city': 'Test city',
+        cls.city = cls.env['res.city'].create({
+            'name': 'Ghotam',
```
pedrobaeza: Is this for avoiding copyright issues with DC, hehe? Or a typo in `Gotham`?

crm_location/tests/test_crm_location.py
```diff
@@ -38,9 +43,9 @@ def test_on_change_city(self):
         lead.location_id = location.id
         lead.on_change_city()
         self.assertEqual(lead.zip, location.name)
-        self.assertEqual(lead.city, location.city)
-        self.assertEqual(lead.state_id, location.state_id)
-        self.assertEqual(lead.country_id, location.country_id)
+        self.assertEqual(lead.city, location.city_id.name)
```
pedrobaeza: This test is not well designed, as we are repeating the implementation in the test itself. We should assert data, not "code". Not sure if you follow me...

What we have to put here is:
```python
self.assertEqual(lead.city, 'Ghotam`)
```
Why? Because on current way, we are making the test dependent of the implementation, and the proof is that you needed to change it when the code has changed.

Please replace all of these occurrences.

pedrobaeza (changes requested):
Other things I have found in the commit history:

- You can squash `[UPD] prefix versions with 8.0` commit with the first commit (which is the module being added in 8.0), as it was a massive commit made by a bot (although through Stéphane's identity) because we didn't put at the beginning fully qualified version.
- There's a commit called "[MIG] Migration Operations:" that is attributed to me with the auxiliary operations I also did through a bot. They should dissappear, as they don't add any valuable information. Remember that the commit "[MIG] Rename manifest files" must be squashed with the migration to version 10.0 commit, and the other one must be removed and possible latter conflicts be solved.
- But not only that, I see that there are translation commits squashed in the same commit. Adjacent translation commits should be squashed together when same author (the original bot or now with Weblate the same translator), but not with other commits.

Tell me if you have any doubt and we meet for solving it.

PR #277, opened 2019-06-03
[ADD] crm_mailchimp

hbrunn:
This module allows to push Odoo partner data to a mailchimp audience.


pedrobaeza:
CRM? Shouldn't this be in OCA/social?

hbrunn:
I considered that, but figured it fits better here - IMHO most mass_mailing_* addons from social should be here actually...

pedrobaeza:
I don't actually (I prefer OCA/social or creating OCA/mass-mailing), but as they are there already, I think we should move this one there. And this is not pluggable to `mail_tracking` modules for same structure?

pedrobaeza:
For reference, we have an old implementation before converting to Mailchimp in https://github.com/OCA/social/tree/8.0/mail_mandrill

hbrunn:
no mail tracking, this addon doesn't care about actual mailings or their results - that could be added of course by someone else.

So given we'll probably be the only people talking here, what name do you suggest for moving this to social? It's neither really a mail thing nor a mass_mailing thing.

pedrobaeza:
Uhm, I see... Why not simply `mailchimp`, with its own section in apss? About the repository, if this is only for dumping general data, indeed OCA/social doesn't fit totally, but I'm seeing mailchimp list that can/should be mailing lists maybe with a special flags, don't you think? Or do you want to keep them isolated and sync them if needed?

hbrunn:
there are two ways to use lists in mailchimp: Either you actually have one list per, ehm, list. In this case, the mail mailing lists map very well. Or you have one big list of people from where mailchimp picks entries according to some criteria like custom fields, tags, interests. That's how my customer uses it, and this maps quite badly to the mass mailing lists.

So this plugin basically pushes a list of mailadresses + metadata to mailchimp for further processing there, and keeps this metadata in Odoo

pedrobaeza:
And that information can't be added to current mass mailing lists? See for example mass_mailing_dynamic_list that also builds the list according a domain that adds "metadata" to the list.

hbrunn:
it can be done, but not in a way that fits what my customer wants. And it's also not sensible for them, because with this workflow, Odoo has nothing at all to do with mailing things, it's just a data source, so I don't see how forcing this into the mass mailings data model improves anything.

mass_mailing_mailchimp simply is another use case.

pedrobaeza:
I understand, so go ahead in the way you consider better.

rafaelbn:
I don't want to make your head dizzy but did you think in OCA/partner-contact and partner_mailchimp as we are adding functionality to the res.partner


> I considered that, but figured it fits better here - IMHO most mass_mailing_* addons from social should be here actually...

:smile: it's not the scope of this PR but crm is not the right repository to mass_mailing_* addons. I agree that "social" is not the best name...

Tested this module functionally :+1: 

---------------
Mailchimp integration
By Therp BV,Odoo Community Association (OCA)

This module allows to push Odoo partner data to a mailchimp audience.

Configuration

To configure this module, you need to:

- go to Settings / Mailchimp / Settings and fill in your mailchimp credentials (this will import audiences)
- go to Settings / Mailchimp / Audiences and review the merge fields in audiences. For every merge field, fill in python code that generates this field’s value (ie: for FNAME, you probably need partner.name)

Usage

- go to some partner as member of the group Mailchimp / User
- on the mailchimp tab, assign as least an audience and possibly groups

Next time you push the list to mailchimp (either manually or via cronjob), this partner will be added or updated in the mailchimp audience.

Known issues / Roadmap

- support tags
- support more mailchimp hooks
- support propagating deletions from Odoo to mailchimp

rafaelbn (approved):

NL66278 (changes requested):
Code looks good and can verify it works in practice.
Two requests:
1. Adding a test for the webhook, and maybe for the name_get on interest would increase coverage and get everything green;
2. Some read only views added to the configuration menu to see the categories / interests retrieved from mailchimp.

hbrunn:
@NL66278 thanks, added coverage. About (2): You see that in the list of audiences, there when you open an audience, you see its merge fields and interest categories, and within the interest categories the available interests.

hbrunn:
PS: Needs more coverage, will be revisited

NL66278 (approved):
:+1: LGTM and working in practice

PR #238, opened 2018-11-13
[11.0][ADD] crm_stage_type

MiquelRForgeFlow:
CRM Stage Type
==============

This module extends the functionality of the CRM adding a type in the stages.

PR #280, opened 2019-06-06
[12.0][MIG] crm_stage_type

MiquelRForgeFlow:
Standard migration of https://github.com/OCA/crm/pull/238.


crm_stage_type/README.rst
```diff
@@ -0,0 +1,54 @@
...
+==============
+
+This module extends the functionality of the CRM adding a type in the stages.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/111/11.0
```
LoisRForgeFlow, line 16: Can you update to readme fragments?

LoisRForgeFlow (approved):
Works good, tested locally. I also saw that the fragments are in place, just not applied to main README so: :+1:

crm_stage_type/README.rst
MiquelRForgeFlow: It already has readme fragments. When merge, it will create a new README.

JordiBForgeFlow (approved):

PR #287, opened 2019-07-08
[12.0][FIX] crm_phonecall: make states editable

oihane:
There is a status called pending that it will never be used because you can not change the status manually. If you take a look, although the status bar is clickable the field is read-only, so when you click on it, it does nothing.

Any comment will be appreciated.

### DIFF
--- a/crm_phonecall/models/crm_phonecall.py
+++ b/crm_phonecall/models/crm_phonecall.py
@@ -54,7 +54,6 @@ class CrmPhonecall(models.Model):
          ('done', 'Held')
          ],
         string='Status',
-        readonly=True,
         track_visibility='onchange',
         default='open',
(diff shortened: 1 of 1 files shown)


crm_phonecall/models/crm_phonecall.py
```diff
@@ -54,7 +54,7 @@ class CrmPhonecall(models.Model):
          ('done', 'Held')
          ],
         string='Status',
-        readonly=True,
+        readonly=False,
```
pedrobaeza: Just remove the line then

oihane: DONE

pedrobaeza (approved):

Daniel-CA (approved):
LGTM

PR #288, opened 2019-07-09
[12.0][FIX] crm_phonecall: when accessing from partner's opportunities active_id is not correct

oihane:
If you access a partner, and from there access opportunities, when you click into calls of the opportunity, it shows calls related to other opportunity, it takes partner id as opportunity id.

### DIFF
--- a/crm_phonecall/views/crm_lead_view.xml
+++ b/crm_phonecall/views/crm_lead_view.xml
@@ -7,10 +7,9 @@
         <field name="inherit_id" ref="crm.crm_case_form_view_leads"/>
         <field name="arch" type="xml">
             <xpath expr="//div[@name='button_box']" position="inside">
-                <button class="oe_inline oe_stat_button" type="action"
-                    name="%(crm_phonecall.crm_case_categ_phone_incoming0)d"
-                    icon="fa-phone"
-                    context="{'default_opportunity_id': active_id, 'search_default_opportunity_id': active_id, 'default_partner_id': partner_id, 'default_duration': 1.0}">
+                <button class="oe_inline oe_stat_button" type="object"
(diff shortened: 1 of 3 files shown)


crm_phonecall/views/crm_lead_view.xml
```diff
@@ -10,7 +10,7 @@
                 <button class="oe_inline oe_stat_button" type="action"
                     name="%(crm_phonecall.crm_case_categ_phone_incoming0)d"
                     icon="fa-phone"
-                    context="{'default_opportunity_id': active_id, 'search_default_opportunity_id': active_id, 'default_partner_id': partner_id, 'default_duration': 1.0}">
+                    context="{'default_opportunity_id': id, 'search_default_opportunity_id': id, 'default_partner_id': partner_id, 'default_duration': 1.0}">
```
pedrobaeza: Is `id` always available in this context? I would say this is more a problem of Odoo than the module itself, with the proper handling of `active_id`...

oihane: I think it is available if there is not id it is because is "new" and then there won't be any phone call.

And of course, the problem comes from Odoo. It isn't the first time I get this same problem. Another way to fix it is by creating a method instead of using an action, what do you think?

pedrobaeza: My fear is that we get an error using `id`, as it's a "magic" field and not always available depending on the web client. Yes, a method would be a safer option.

oihane: Done using method. :wink:

pedrobaeza (approved):

anajuaristi (approved):

PR #289, opened 2019-07-30
[12.0][crm_lead_code] Update crm_lead_view.xml

dammaksalma:


dammaksalma:
Add 'name' field to search view on 'view_crm_case_leads_filter'

rafaelbn:
This PR is the same of https://github.com/OCA/crm/pull/290

keshrath (commented):
I'm not sure what you're trying to improve with this PR. There is already a name field in this SerachView and it is extended to search for code as well. 

https://github.com/odoo/odoo/blob/12.0/addons/crm/views/crm_lead_views.xml#L388

oihane (approved):
LGTM

crm_lead_code/views/crm_lead_view.xml
```diff
@@ -28,6 +28,9 @@
             <field name="model">crm.lead</field>
             <field name="inherit_id" ref="crm.view_crm_case_leads_filter" />
             <field name="arch" type="xml">
+                <xpath expr="//field[@name='user_id']" position="before">
+                    <field name="name"/>
```
sbejaoui, line 32: Field name is already in the search view, why you need to add it twice?

sbejaoui (changes requested):

PR #291, opened 2019-08-16
[FIX] crm_phonecall button size

ivantodorovich:
This fixes the button size in partner and opportunities

Before:
image

After:
image

### DIFF
--- a/crm_phonecall/views/res_partner_view.xml
+++ b/crm_phonecall/views/res_partner_view.xml
@@ -8,14 +8,14 @@
         <field name="groups_id" eval="[(4, ref('sales_team.group_sale_manager')), (4, ref('base.group_partner_manager'))]"/>
         <field eval="18" name="priority"/>
         <field name="arch" type="xml">
-            <xpath expr="//div[@name='button_box']" position="inside">
-                <button class="oe_inline oe_stat_button" type="action"
+            <button name="toggle_active" position="before">
+                <button class="oe_stat_button" type="action"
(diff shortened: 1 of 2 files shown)


NL66278 (approved):
:+1: LGTM

crm_phonecall/views/res_partner_view.xml
```diff
@@ -8,14 +8,14 @@
         <field name="groups_id" eval="[(4, ref('sales_team.group_sale_manager')), (4, ref('base.group_partner_manager'))]"/>
         <field eval="18" name="priority"/>
         <field name="arch" type="xml">
-            <xpath expr="//div[@name='button_box']" position="inside">
```
pedrobaeza, line 11: Please don't change this. It should be inside the div, not before another button

pedrobaeza (changes requested):

crm_phonecall/views/res_partner_view.xml
ivantodorovich: I did this because otherwise the button placed after the "Active" button. And it doesn't look good. All other buttons in odoo are added like that.

ie: the crm Opportunities button:
image


Changing it like this have the same effect, only it keeps the toggle_active button at the end, which is the standard in odoo

pedrobaeza: OK, thanks for the extra explanations. Seems fair. In v13 there's no more active smart-button, so we will need to change it again, hehe.

pedrobaeza (approved):

PR #294, opened 2019-09-16
[12.0][ADD] crm_won_reason: funcionality to set won reasons

dsidorenkovas-rgbconsulting:
This module adds won reason funcionality to leads (same as lost reason).

A lead can be marked as won through a button or an action. Either way,
both methods open a wizard that allows specifying the won reason.
The reason, when set, is shown on the form view of the lead.

### DIFF
--- a/crm_won_reason/__init__.py
+++ b/crm_won_reason/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import wizard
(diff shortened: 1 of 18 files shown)


lfreeke (approved):
Functional test :+1: 

I think you need to add a test to update the code coverage.

PR #302, opened 2020-01-06
[13.0] Add module crm_stage_probability

grindtildeath:
This module restore the CRM feature from Odoo <= 12.0 with lead probability
according to its stage.


crm_stage_probability/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2019 Camptocamp SA
```
Tonow-c2c:
```suggestion
# Copyright 2020 Camptocamp SA
```

crm_stage_probability/models/crm_lead.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2019 Camptocamp SA
```
Tonow-c2c:
```suggestion
# Copyright 2020 Camptocamp SA
```

crm_stage_probability/models/crm_lead.py
```diff
@@ -0,0 +1,77 @@
...
+    _inherit = "crm.lead"
+
+    def _default_probability(self):
+        if "default_stage_id" in self._context:
+            stage_id = self._context.get("default_stage_id")
+        else:
+            stage_id = self._default_stage_id()
+        if stage_id:
+            return self.env["crm.stage"].browse(stage_id).probability
+        return 10
```
Tonow-c2c, line 17: I will do only one return with one variable like

```python
    def _default_probability(self):
        probability = 10
        if "default_stage_id" in self._context:
            stage_id = self._context.get("default_stage_id")
        else:
            stage_id = self._default_stage_id()
        if stage_id:
            probability = self.env["crm.stage"].browse(stage_id).probability
        return probability
```

crm_stage_probability/models/crm_stage.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2019 Camptocamp SA
```
Tonow-c2c:
```suggestion
# Copyright 2020 Camptocamp SA
```

crm_stage_probability/wizard/crm_lead_stage_probability_update.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2019 Camptocamp SA
```
Tonow-c2c:
```suggestion
# Copyright 2020 Camptocamp SA
```

Tonow-c2c (approved):

crm_stage_probability/models/crm_lead.py
grindtildeath: It's actually a copy-paste from https://github.com/odoo/odoo/blob/12.0/addons/crm/models/crm_lead.py#L56-L63 so :man_shrugging:

p-tombez (approved):

fclementic2c (approved):

crm_stage_probability/models/crm_lead.py
gurneyalex: Can you add Odoo SA as one of the authors of the module, if you copy pasted code from their repo ?

gurneyalex: (https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility -> in case people are wondering, it is allowed to licence this module under AGPL, even with LGPL code copied)

gurneyalex (approved):

PR #312, opened 2020-03-19
[13.0][MIG] crm_location

newtratip:
Standard Migration.
Peek 2020-03-25 17-24


rousseldenis (approved):
Code review

crm_location/tests/test_crm_location.py
```diff
@@ -0,0 +1,52 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0
+
+from odoo.tests import common
+
+
+class TestCrmLocation(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestCrmLocation, cls).setUpClass()
+        cls.country = cls.env["res.country"].create({"name": "Test country"})
```
simahawk, line 13:
```suggestion
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
        cls.country = cls.env["res.country"].create({"name": "Test country"})
```
BTW can't we simply use the existing records?

crm_location/models/crm_lead.py
```diff
@@ -0,0 +1,30 @@
...
+class CrmLead(models.Model):
+    _inherit = "crm.lead"
+
+    @api.onchange("location_id")
+    def on_change_city(self):
+        if self.location_id:
+            self.zip = self.location_id.name
+            self.city = self.location_id.city_id.name
+            self.state_id = self.location_id.city_id.state_id
+            self.country_id = self.location_id.city_id.country_id
```
simahawk:
```suggestion
            self.update({
                "zip": self.location_id.name,
                "city": self.location_id.city_id.name,
                "state_id": self.location_id.city_id.state_id.id,
                "country_id": self.location_id.city_id.country_id.id,
            })
```

crm_location/tests/test_crm_location.py
newtratip: @simahawk Why do you use `tracking_disable=True` in the context ? Please suggest me.

can't we simply use the existing records? >> Do you mean use the existing records replace of creating new record ?

crm_location/models/crm_lead.py
newtratip: Edited, Thanks.

crm_location/tests/test_crm_location.py
newtratip: fixed.

simahawk: because here we create of records that are probably not required and this add a lot of writes which slow down tests (consider we have tons of tests in all repos).
We could reuse existing demo data and change only the fields we need it.
Using tracking_disable makes sure we shut down tracking machinery which is useless and brings a lot of not required overhead.

simahawk (approved):

PR #314, opened 2020-03-25
new module base_newsletter_sendy

jans23:
This module allows ecommerce customers to subscribe a newsletter which is provided by [Sendy](https://sendy.co/). Sendy is a self-hosted newsletter system independent of Odoo.


base_newsletter_sendy/models/res_partner.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @api.model
+    def create(self, vals):
+        res = super(Partner, self).create(vals)
+        if vals.get('newsletter_sendy') and vals.get('email'):
+            res.change_newsletter_sendy(vals.get('email'), True)
+        return res
+
+    @api.multi
+    def copy(self, default=None):
```
rousseldenis, line 86: You can do that in field attributes:
`copy=False`

base_newsletter_sendy/migrations/11.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# Copyright 2018 - Nicolas JEUDY
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+#_xmlid_renames = [
```
rousseldenis, line 7: Remove this if unused

rousseldenis (changes requested):
Some remarks.

PR #315, opened 2020-04-03
[13.0][MIG] crm_claim

flachica:
I closed #311 and open this to facilitate the revision

PR #311, opened 2020-03-18
[13.0][MIG] crm_claim

flachica:

PR #316, opened 2020-04-04
[13.0][MIG] crm_claim

flachica:
In substitution of #315 for not having done squash which substitutes #311 for having conflicts


dreispt (approved):
IMO this commit history cleansing is low value bike shedding.
But that's me, so follow the crowd.

traviswaelbro:
> IMO this commit history cleansing is low value bike shedding.
> But that's me, so follow the crowd.

This PR is 21 commits, nearly half of which are readme/translation commits. I think it's worth an extra 5 minutes to keep a cleaner commit history (and in this case have nearly half as many commits as otherwise).

crm_claim/README.rst
```diff
@@ -0,0 +1,104 @@
...
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+This module is part of the `OCA/crm <https://github.com/OCA/crm/tree/12.0/crm_claim>`_ project on GitHub.
```
mamcode: Please change URL to 13.0. I think that you should regenerate the README for this and also the list of collaborators will also be updated... You can also not send the readme and I think the bot will generate it automatically when merging the branches.

mamcode (changes requested):
Hi Fernando, Thanks for this work, looks good to me, only I see necessary a minor change...

mamcode:
> This PR is 21 commits, nearly half of which are readme/translation commits. I think it's worth an extra 5 minutes to keep a cleaner commit history (and in this case have nearly half as many commits as otherwise).

+1

flachica:
@traviswaelbro @mamcode History rewrited, minor change pushed. I think that readme is regenerated on merge with 13.0 but I does this. I think as @dreispt, rewrite history is low value but I follow the crowd too

mamcode (approved):

simahawk:
@flachica next time please keep module name in migration commit, thanks (more important than rewriting history IMO) :wink:

PR #322, opened 2020-04-27
[13.0][IMP] crm_phonecall: Add field direction in calls and improve views.

mamcode:
**Description of the issue/feature this PR addresses:**

* A better organization of layout in form view of calls.
* Add new field direction to indicate if the call is outgoing or incoming. Also add a filter for it in search view.
* Add a new filters for field date in search view.
* Add a new group (by) for field state in search view.
* Set to visible the field state in tree view.

screenshot-localhost_8099-2020 04 27-07_44_09

screenshot-localhost_8099-2020 04 27-07_47_23

### DIFF
--- a/crm_phonecall/models/res_config_settings.py
+++ b/crm_phonecall/models/res_config_settings.py
@@ -1,0 +1,12 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    group_show_form_view = fields.Boolean(
+        "Enable form view for phone calls",
+        help="By default form is disabled for calls, with this group it is enabled.",
+        implied_group="crm_phonecall.group_show_form_view",
+        default=True,
+    )
--- a/crm_phonecall/views/crm_phonecall_view.xml
+++ b/crm_phonecall/views/crm_phonecall_view.xml
@@ -311,29 +311,29 @@
                 <group expand="0" string="Group By">
                     <filter
                         string="Partner"
-                        name="partner"
+                        name="groupby_partner"
                         context="{'group_by':'partner_id'}"
                     />
                     <filter
                         string="Responsible"
-                        name="responsible"
+                        name="groupby_responsible"
                         context="{'group_by':'user_id'}"
                     />
                     <filter
                         string="Creation"
-                        name="creation"
+                        name="groupby_creation"
                         help="Creation Date"
                         context="{'group_by':'create_date'}"
                     />
                     <filter
                         string="Month"
-                        name="month"
+                        name="groupby_month"
                         context="{'group_by':'date'}"
                         help="Calls Date by Month"
                     />
                     <filter
                         string="State"
-                        name="state"
+                        name="groupby_state"
                         context="{'group_by':'state'}"
                         help="Calls by status"
                     />
--- a/crm_phonecall/views/crm_phonecall_view.xml
+++ b/crm_phonecall/views/crm_phonecall_view.xml
@@ -191,6 +191,7 @@
                     context="{'default_groups_ref': ['base.group_user', 'base.group_partner_manager', 'sales_team.group_sale_salesman']}"
                 />
                 <field name="tag_ids" widget="many2many_tags" invisible="1" />
+                <field name="direction" />
                 <field name="state" />
                 <field name="create_date" invisible="1" />
                 <field name="opportunity_id" invisible="1" />
--- a/crm_phonecall/views/crm_phonecall_view.xml
+++ b/crm_phonecall/views/crm_phonecall_view.xml
@@ -120,45 +120,52 @@
                             <field name="partner_phone" />
                         </h2>
                     </div>
-                    <group col="4">
-                        <field name="date" />
-                        <field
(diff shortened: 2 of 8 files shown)


NL66278 (approved):
:+1: LGTM

rousseldenis (approved):
Code review

skeller1:
creating first logged call: form opens
creating second record: tree opens in inline mode top
Maybe we could remove the editable="top" from the tree view?
Defined [here](https://github.com/OCA/crm/blob/a6906b2fd684393271d7f29276aee607c0585fe0/crm_phonecall/views/crm_phonecall_view.xml#L183)

I have to remove the state group by if I want to add a second logged call for a opportunity (Create button is not available) 
improve_call

:+1: The form changes: LGTM

Thank you for improving this addon

ivantodorovich (approved):

ivantodorovich:
> creating first logged call: form opens
creating second record: tree opens in inline mode top
Maybe we could remove the editable="top" from the tree view?


I would be in favour of removing the  `editable="top"` in the tree view.
@NL66278 @rousseldenis what do you think?

NL66278:
I am not a great fan of editable trees, so I have no problem with removing the `editable="top"`.

pedrobaeza:
I'm the contrary: editable trees makes your work very agile, so please don't remove it.

ivantodorovich:
> I'm the contrary: editable trees makes your work very agile, so please don't remove it.

The issue is that it's really difficult for the user to get to the form view (I'd say impossible), and not all fields are displayed in the tree view

pedrobaeza:
Which fields are not shown? Why not include them in tree? I think the number is still manageable. If not, when grouped, it goes to form view. You can also propose both menus: a fast-edit menu with editable="bottom" and another without, or a configuration option (like the one in sales > packaging that disables the inline editability, but if you don't need packaging, you can still use the "simple" tree).

ivantodorovich:
> Which fields are not shown? Why not include them in tree? I think the number is still manageable. If not, when grouped, it goes to form view. You can also propose both menus: a fast-edit menu with editable="bottom" and another without, or a configuration option (like the one in sales > packaging that disables the inline editability, but if you don't need packaging, you can still use the "simple" tree).
 
tags, sales team, opportunity, priority, direction

I like the idea of adding a configuration option as it covers all preferences :+1:

mamcode:
I have extended the tree view of calls to remove `editable=top` but this view is only enabled for users with a new group `crm_phonecall.group_show_form_view`.
screenshot-localhost_8099-2020 05 04-12_53_01

Please review again @ivantodorovich @pedrobaeza @skeller1 @NL66278 @rousseldenis

crm_phonecall/models/res_config_settings.py
```diff
@@ -0,0 +1,12 @@
...
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    group_show_form_view = fields.Boolean(
+        "Enable form view for phone calls",
+        help="By default form is disabled for calls, with this group it is enabled.",
+        implied_group="crm_phonecall.group_show_form_view",
+        default=True,
+    )
```
ivantodorovich, line 12: You need to add this to the res.config view as well

You can add a new section "Phonecalls" there
image

ivantodorovich (changes requested):

mamcode:
@ivantodorovich Done.

skeller1 (approved):

skeller1 (commented):
The changes look great. thank you.
From the perspective of a CRM User:
1. Go to to my opportunity or lead
2. Want to call the lead/partner, but before have a look at the calls hold in the past to be up to date
3. Clicking on Calls (Stats Button)
4. Have a list entry with Hold records
5. I have to remove the state filter or toggle the group to see some information
6. Removing the filter or expanding the group in the list view could be very frustrating doing it for each CRM Call history lookup in daily business


So maybe we could expand the groups by default or order them  by date (last, youngest call on top)

crm_phonecall/views/crm_phonecall_view.xml
```diff
@@ -296,6 +331,12 @@
                         context="{'group_by':'date'}"
                         help="Calls Date by Month"
                     />
+                    <filter
+                        string="State"
+                        name="state"
```
ivantodorovich:
```suggestion
                        name="groupby_state"
```

Can you rename the others group by as well, for consistancy?

I think that's what's causing the issue reportred by @skeller1 
There's a context key in the action: `'search_default_state': 'done'`.
The intention is to apply a default state filter, but it's also triggering the group by filter because it has the same name.

ivantodorovich (changes requested):

crm_phonecall/views/crm_phonecall_view.xml
mamcode: I have sent a commit fixing this @ivantodorovich

ivantodorovich (approved):

ivantodorovich:
Is it ok to merge this one? @NL66278 @rousseldenis ?

NL66278:
As for me, this can be merged.

PR #323, opened 2020-05-02
[13.0][MIG] crm_stage_type: Migration to 13.0 

skeller1:
**Migration related changes:**
- [x] Changed category from `CRM` to `Sales`
- [x] removed api.multi
- [x] removed oldname type for lead_type field
- [x] lead and opportunity using the same form (less xml, changing attributes of stage_id field to display it in both contexts (lead and opportunity)

**Improvements:**
- [x] moved `lead_type` field in crm.stage form before `is_won` field
- [x] added tests (different stages for lead and opportunity, tests for `_read_group_stage_ids`)
- [x] improved `_stage_find` (using super class instead of Copy&Paste Odoo original implementation)

**Problems, open questions:**
- [x] **SOLVED**: possible dependency (incomatibility) between `crm_stage_type` and `crm_stage_probability` addons (during opportunity conversion in Odoo 12 probability has been set by this crm_stage_type addon: https://github.com/OCA/crm/blob/8592a3d52ca6f9ad92d0760d810c58f5d7175b5c/crm_stage_type/models/crm_lead.py#L82 **(SOLVED: probability on converted lead is recalculated automatically after each create/write)**


skeller1 (commented):
Ready for review

crm_stage_type/__manifest__.py
```diff
@@ -1,16 +1,16 @@
...
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "CRM Stage Type",
     "summary": "Add type in the lead and opportunity stages",
-    "version": "12.0.1.0.0",
-    "category": "CRM",
+    "version": "13.0.1.0.0",
+    "category": "Customer Relationship Management",
```
rousseldenis: @skeller1 You need to follow Odoo standard categories in order to appear in OCA shop right category. In CRM, it's 'Sales'.

skeller1: Changed the category to `Sales` and changed tests (using SavepointCase)

**Question: Why all the other OCA/crm addons have a custom category?**

rousseldenis: > Question: Why all the other OCA/crm addons have a custom category?

Everybody do not pay attention to that :smile:

rousseldenis (approved):
Code review

crm_stage_type/__manifest__.py
skeller1: Okay, for further developments: only use these default categories, correct?

https://github.com/odoo/odoo/blob/13.0/odoo/addons/base/data/ir_module_category_data.xml

thank you

HaraldPanten (approved):
Functional review. LGTM 👍

crm_stage_type/models/crm_lead.py
```diff
@@ -0,0 +1,62 @@
...
+        if ctx_type:
+            search_domain += [("lead_type", "in", [ctx_type, "both"])]
+        stage_ids = stages._search(
+            search_domain, order=order, access_rights_uid=SUPERUSER_ID
+        )
+        return stages.browse(stage_ids)
+
+    def _stage_find(self, team_id=False, domain=None, order="sequence"):
+        # check whether we should try to add a condition on type
+        if domain and not any(
```
skeller1: @HaraldPanten: I don't know if we introduce a bug here (domain could be None)

skeller1 (commented):
Needs a fix for calling `_stage_find` without domain

crm_stage_type/models/crm_lead.py
HaraldPanten: hhhmmm.... @ValentinVinagre what do you think? might a bug be introduced here as @skeller1 is suggesting? Thx!

skeller1: crm.lead model is calling _stage_find() without a domain
[see here in 13.0](https://github.com/odoo/odoo/blob/2100f07cf2eea3c813ea1ee702d25b0beb3a1e8e/addons/crm/models/crm_lead.py#L879)
[see here in 12.0 ](https://github.com/odoo/odoo/blob/198ab837e466f3155ea1c3b35811b981b1c1e537/addons/crm/models/crm_lead.py#L777)

The current crm_stage_type 12.0 implementation should raise an error if we pass no domain:
`Traceback (most recent call last):
TypeError: 'NoneType' object is not iterable
`
Fixed it here defining domain as an empty list (iteratable) if it is None

ValentinVinagre: Hi @skeller1 you add `domain = domain or []`?
This fixes the problem, I think it is a correct solution.
Another solution could be:
 `[term for term in (domain or []) if len(term) == 3 if term[0] == "lead_type"]` 
but since that variable can be used later I think the applied solution is correct.
👍

crm_stage_type/models/crm_lead.py
```diff
@@ -0,0 +1,63 @@
...
+    def merge_opportunity(self, user_id=False, team_id=False):
+        opportunities_head = super(Lead, self).merge_opportunity(user_id, team_id)
+        if opportunities_head.team_id:
+            team_stage_ids = self.env["crm.stage"].search(
+                [
+                    ("team_id", "in", [opportunities_head.team_id.id, False]),
+                    ("lead_type", "in", [opportunities_head.type, "both"]),
+                ],
+                order="sequence",
+            )
```
ValentinVinagre, line 49: If you always use one record, you can add limit = 1, and then in the `team_stage_ids [0]` you can delete `[0]` more faster 👍

crm_stage_type/models/crm_lead.py
```diff
@@ -0,0 +1,63 @@
...
+        stage_ids = stages._search(
+            search_domain, order=order, access_rights_uid=SUPERUSER_ID
+        )
+        return stages.browse(stage_ids)
+
+    def _stage_find(self, team_id=False, domain=None, order="sequence"):
+        # check whether we should try to add a condition on type
+        domain = domain or []
+        if not any(
+            [term for term in domain if len(term) == 3 if term[0] == "lead_type"]
```
ValentinVinagre: why use 2 if's? you could use 'and'.
`[term for term in domain if len(term) == 3 and term[0] == "lead_type"]`

ValentinVinagre (commented):
Some comments, but nice work 👍

crm_stage_type/models/crm_lead.py
skeller1: This implementation is needed if there's a stage used by many teams.
If the stage not exists for this team_id, then the first found record is set as new opportunity state. We need this becaus we could also define stages without filled team_id

skeller1:
Have to fix the old tests migrated from 12.0
They are using asserTrue to compare the stage of created leads/opportunities
This is wrong, fixing it, set the PR to draft

skeller1:
Fixed using assertTrue instead of assertEqual, also changed it for crm_stage_type 12.0 in #328 
Thank you for all the help

There's still one open problem: The compatibility with crm_stage_probability

HaraldPanten (approved):
Hi @skeller1 ,

Functionally is working fine, but I don't understand why is this module incompatible with crm_stage_probability. What do you mean?

Thanks!

ValentinVinagre:
Hi @skeller1 ,
Maybe you can check if 'is_stage_probability' or 'stage_probability' fields exist. If any of these fields exist, don't write the probability.

IMO this solution solves the conflict.

skeller1:
> Hi @skeller1 ,
> Maybe you can check if 'is_stage_probability' or 'stage_probability' fields exist. If any of these fields exist, don't write the probability.
> 
> IMO this solution solves the conflict.

@HaraldPanten , @ValentinVinagre: I've looked in the code and found [this](https://github.com/odoo/odoo/blob/bd282eea08e0b4814d327ec36c0f743dd6cd9914/addons/crm/models/crm_lead.py#L438) in lead model.
AFAIU this should correct the probability after changing the stage in `_convert_opportunity_data`
So we don't need to change anything.

Thank you for your support

rafaelbn (approved):

PR #328, opened 2020-05-12
[12.0][IMP] crm_stage_type: add missing tests

skeller1:
**Add missing tests for crm_stage_type addon:**

- [x] _read_group_stage_ids
- [x] improve code coverage for already existing tests

### DIFF
--- a/crm_stage_type/tests/test_crm_stage_type.py
+++ b/crm_stage_type/tests/test_crm_stage_type.py
@@ -1,72 +1,112 @@
 # Copyright 2018 Eficent Business and IT Consulting Services, S.L.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo.tests.common import TransactionCase
+from odoo.tests.common import SavepointCase
 
 
-class TestCrmStageType(TransactionCase):
-    def setUp(self):
-        super(TestCrmStageType, self).setUp()
-
-        self.crm_salemanager = self.env['res.users'].create({
-            'company_id': self.env.ref("base.main_company").id,
+class TestCrmStageType(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        context = dict(cls.env.context, tracking_disable=True, no_reset_password=True)
+        cls.env = cls.env(context=context)
+        cls.crm_salemanager = cls.env['res.users'].create({
+            'company_id': cls.env.ref("base.main_company").id,
             'name': "Crm Sales manager",
             'login': "csm",
             'email': "crmmanager@yourcompany.com",
-            'groups_id': [(6, 0, [self.ref('sales_team.group_sale_manager')])]
+            'groups_id': [(6, 0, [cls.env.ref('sales_team.group_sale_manager').id])]
         })
+        # archiving default odoo stages is not possible, so change their lead_type
+        # and use Won stage for both types
+        # Opportunity stages: New => Qualified => Proposition => Won [both]
+        # Lead stages: A => B => Won [both]
+        cls.env.ref("crm.stage_lead1").write({'lead_type': "opportunity"})
+        cls.env.ref("crm.stage_lead2").write({'lead_type': "opportunity"})
+        cls.env.ref("crm.stage_lead3").write({'lead_type': "opportunity"})
 
-    def test_find_stage(self):
-        lead = self.env['crm.lead'].create({
+        Stage = cls.env['crm.stage']
+
+        cls.lead_stage_a = Stage.create(
+            {'name': "A", 'lead_type': "lead", 'sequence': 4}
+        )
+        cls.lead_stage_b = Stage.create(
+            {'name': "B", 'lead_type': "lead", 'sequence': 5}
+        )
+        # freeze stage ids for tests
+        cls.stages = [
+            cls.env.ref("crm.stage_lead1").id,
+            cls.env.ref("crm.stage_lead2").id,
+            cls.env.ref("crm.stage_lead3").id,
+            cls.env.ref("crm.stage_lead4").id,
+            cls.lead_stage_a.id,
+            cls.lead_stage_b.id
+        ]
+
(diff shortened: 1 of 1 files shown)


crm_stage_type/tests/test_crm_stage_type.py
```diff
@@ -16,26 +16,54 @@ def setUp(self):
             'groups_id': [(6, 0, [self.ref('sales_team.group_sale_manager')])]
         })
```
simahawk, line 19: as you are improving tests, would you mind replacing TransactionCase with Savepoint case and replace setUp like this?
```py
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True, no_reset_password=True))
        cls.crm_salemanager = cls.env['res.users'].create({
            'company_id': cls.env.ref("base.main_company").id,
            'name': "Crm Sales manager",
            'login': "csm",
            'email': "crmmanager@yourcompany.com",
            'groups_id': [(6, 0, [cls.ref('sales_team.group_sale_manager')])]
        })

File: crm_stage_type/tests/test_crm_stage_type.py [hunk 2]
```diff
@@ -16,26 +16,54 @@ def setUp(self):
...
-        self.assertTrue(lead.stage_id.lead_type, 'both')
-        stage_id = lead._stage_find(domain=[('probability', '=', 100.0)])
-        self.assertEqual(stage_id, lead.stage_id,
-                         "Stage of opportunity is incorrect!")
-        self.assertTrue(stage_id.lead_type, 'both')
+
+    def test_read_group_stage_ids(self):
+        # test returned stages for lead
+        groups_lead = self.lead._read_group_stage_ids(
+            self.env['crm.stage'].search([]), None, "sequence"
```
simahawk: I would freeze the ids you are reading (you already know them) because if something changes in demo data, this test will be broken

crm_stage_type/tests/test_crm_stage_type.py
```diff
@@ -16,26 +16,54 @@ def setUp(self):
...
+            groups_lead.ids,
+            [
+                self.lead_stage_a.id,
+                self.lead_stage_b.id,
+                self.env.ref('crm.stage_lead4').id,
+            ],
+        )
+
+    def test_find_stage(self):
+        self.assertTrue(self.lead.stage_id, self.lead_stage_a)
```
simahawk: this assert seems useless as you create them on setup

crm_stage_type/tests/test_crm_stage_type.py
```diff
@@ -16,26 +16,54 @@ def setUp(self):
...
+        self.lead.convert_opportunity(self.env.ref("base.res_partner_2").id)
+        self.assertLessEqual(
```
simahawk: less?

simahawk (changes requested):

crm_stage_type/tests/test_crm_stage_type.py
skeller1: Changed it to SavepointCase

crm_stage_type/tests/test_crm_stage_type.py
skeller1: I freezed the ids in the setUp method.

crm_stage_type/tests/test_crm_stage_type.py
skeller1: That's the test to be sure that the lead (type=lead) gets the correct default stage (Stage name A)

skeller1 (commented):
Did all the changes, thank you for checking this :+1:

crm_stage_type/tests/test_crm_stage_type.py
skeller1: Removed this test

crm_stage_type/tests/test_crm_stage_type.py
simahawk: but then you should use assertEqual, here you are testing only that stage_id is valued

skeller1 (commented):
Fixed the test, thank you, but then the current tests (see [here](https://github.com/OCA/crm/blob/d56b6c6c7875e86ef1b1e9a214a8f3774d99bcae/crm_stage_type/tests/test_crm_stage_type.py#L60) using always assertTrue)
So I have to change this, thank you for your hints

crm_stage_type/tests/test_crm_stage_type.py
```diff
@@ -1,41 +1,72 @@
...
-            'groups_id': [(6, 0, [self.ref('sales_team.group_sale_manager')])]
+            'groups_id': [(6, 0, [cls.env.ref('sales_team.group_sale_manager').id])]
         })
 
-    def test_find_stage(self):
-        lead = self.env['crm.lead'].create({
+        Stage = cls.env['crm.stage']
+        # archiving default odoo stages is not possible, so change their lead_type
+        # and use Won stage for both types
+        Stage.browse([1, 2, 3]).write({'lead_type': "opportunity"})
```
simahawk: :fire: do not harcode ids, collect them w/ ref or search

crm_stage_type/tests/test_crm_stage_type.py
```diff
@@ -1,41 +1,72 @@
...
+        # Opportunity: New => Qualified => Proposition => Won [both]
+        # Lead: A => B => Won [both]
+        cls.lead_stage_a = Stage.create(
+            {'name': "A", 'lead_type': "lead", 'sequence': 4}
+        )
+        cls.lead_stage_b = Stage.create(
+            {'name': "B", 'lead_type': "lead", 'sequence': 5}
+        )
+        # freeze stages for tests
+        cls.stages = [1, 2, 3, 4, cls.lead_stage_a.id, cls.lead_stage_b.id]
```
simahawk: :fire:

simahawk (approved):

rafaelbn (approved):

PR #347, opened 2020-07-28
Fix Case of Domain is None

gdgellatly:
In certain cases - such as when a stage is undefined, _stage_find can receive None as the domain argument. This causes a traceback during search. 

This Pull request first checks for domains existence before iteration.


HaraldPanten (approved):
Small FIX. Seems to be OK

rousseldenis (approved):
Obvious!

ivantodorovich (approved):

mamcode (approved):

ivantodorovich (approved):

PR #348, opened 2020-07-29
[12.0][ADD] crm_sale_lead_tag_sync : Sync CRM Lead tags with Sale Order tags

ivantodorovich:
Synchronize tags between Leads and Sale Orders

---

Edit the tags of any lead with a linked sale order, and you'll find the
tags automatically synchronized to the sale order.

You can also edit the tags of any sale order linked to an opportunity,
and you'll find the tags automatically synchronized to the opportunity.

### DIFF
--- a/crm_sale_lead_tag_sync/models/crm_lead.py
+++ b/crm_sale_lead_tag_sync/models/crm_lead.py
@@ -1,0 +1,21 @@
+# Copyright 2020 Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class CrmLead(models.Model):
+    _inherit = 'crm.lead'
+
+    def _sync_tag_ids(self):
+        """ Synchronizes tag_ids to the sale orders. """
+        for rec in self:
+            for order in rec.order_ids:
+                if order.tag_ids != rec.tag_ids:
+                    order.tag_ids = rec.tag_ids
+
+    def write(self, vals):
+        res = super().write(vals)
+        if 'tag_ids' in vals:
+            self._sync_tag_ids()
+        return res
--- a/crm_sale_lead_tag_sync/__init__.py
+++ b/crm_sale_lead_tag_sync/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/crm_sale_lead_tag_sync/__manifest__.py
+++ b/crm_sale_lead_tag_sync/__manifest__.py
@@ -1,0 +1,17 @@
+# Copyright 2020 Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'CRM Lead Tags Sync',
+    'summary': 'Sync CRM Lead tags with Sale Order tags',
+    'category': 'Customer Relationship Management',
+    'version': '12.0.1.0.0',
+    'author': 'Druidoo, '
+              'Moka Tourisme, '
+              'Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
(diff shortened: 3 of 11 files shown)


HaraldPanten (approved):
Works as expected. Functional review 👍

marcelsavegnago (approved):
Functional review ok

bodedra (approved):

rafaelbn:
@ivantodorovich are you going to add the test for maintaining coverage?

ivantodorovich:
> @ivantodorovich are you going to add the test for maintaining coverage?

all green now

rafaelbn:
Thank you @ivantodorovich .

I think the word `sale` should be in the name of the module, like crm_sale_lead_tag_sync (to be really clear).

What do you think?

crm_lead_tag_sync/models/crm_lead.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2020 Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class CrmLead(models.Model):
+    _inherit = 'crm.lead'
+
+    @api.constrains('tag_ids')
```
rousseldenis: You cannot use constrains to do that...

If you want to use something more readable than overriding write(), use inverse property on tag_ids field.

rousseldenis (changes requested):
Please don't use constrains.

crm_lead_tag_sync/models/crm_lead.py
ivantodorovich: I get that it's not really a constraint and it's conceptually wrong, but it gets the job done. I've just changed it anyways.
Out of curiosity, is there a technical reason not to do it?

ivantodorovich:
> Thank you @ivantodorovich .
> 
> I think the word `sale` should be in the name of the module, like crm_sale_lead_tag_sync (to be really clear).
> 
> What do you think?

Agree! I've just renamed it

crm_lead_tag_sync/models/crm_lead.py
rousseldenis: You are not aware of what implementation Odoo's ORM behind constrains(). The purpose is not to write on records. Odoo can change it (maybe invalidate cache or ....) and your code will not work anymore.

crm_sale_lead_tag_sync/models/crm_lead.py
```diff
@@ -0,0 +1,21 @@
...
+    def _sync_tag_ids(self):
+        """ Synchronizes tag_ids to the sale orders. """
+        for rec in self:
+            for order in rec.order_ids:
+                if order.tag_ids != rec.tag_ids:
+                    order.tag_ids = rec.tag_ids
+
+    def write(self, vals):
+        res = super().write(vals)
+        if 'tag_ids' in vals:
```
rousseldenis, line 19: As I said, you can use inverse as it is more 'readable'.

```
tag_ids = fields(inverse=_inverse_tag_ids)

def _inverse_tag_ids(self):
    self._sync_tag_ids()
```

But nitpicking

ivantodorovich: I don't like to use inverse because it decreases compatibility with other modules

rousseldenis: > I don't like to use inverse because it decreases compatibility with other modules

Why ? You can name your function _inverse_tag_ids_tag_sync() if you want

ivantodorovich: We're adding an inverse method to a field that already exist, and you can only have one inverse per field.
I know it's unlikely, but if some other module tries to do the same, only of them will work

It would be different if the field was created by our module, because in that case we'd own the field and any other module wanting to extend it can simply overload the inverse method.

PR #349, opened 2020-08-07
crm_phonecall: Phonecall statusbar not clickable

flachica:
The Phonecall form's statusbar is intended to be clickable, but Odoo 13 is apparently no longer compatible with clickable="True". Instead, either clickable="1" or options="{'clickable': 1}" must be used. Odoo core code uses the latter, so the Claim form has been updated to match core convention.


HaraldPanten (approved):
Functional and code review

marcelsavegnago (approved):

NL66278 (approved):
:+1: LGTM

PR #357, opened 2020-10-22
14.0 migration crm phonecalls

redapureskill:
I have migrated the 13 version of crm_phonecalls to the 14.


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #365, opened 2020-11-06
[14.0][MIG] crm_lead_code: Migration to 14.0

newtratip:
Standard Migration, Ready for review.


bodedra (approved):
Functional review. :+1:

HaraldPanten (approved):

rousseldenis (approved):
Code review

newtratip:
@rousseldenis, Can this merge ?

rousseldenis:
@HaraldPanten Nice merge power :-) 

FYI, when using nobump, package won't be marked as release package on pypi. If you want to do it, use patch (even if it will update module version to X.X.X.X.1)

HaraldPanten:
> @HaraldPanten Nice merge power :-)
> 
> FYI, when using nobump, package won't be marked as release package on pypi. If you want to do it, use patch (even if it will update module version to X.X.X.X.1)


Really? I thought that patch wouldn't be suitable for these [MIG] cases because of the "module version to X.X.X.X.1" changes.

I also see that [MIG] to V13 was done by nobump instead of patch. @rousseldenis

rousseldenis:
> > @HaraldPanten Nice merge power :-)
> > FYI, when using nobump, package won't be marked as release package on pypi. If you want to do it, use patch (even if it will update module version to X.X.X.X.1)
> 
> Really? I thought that patch wouldn't be suitable for these [MIG] cases because of the "module version to X.X.X.X.1" changes.
> 
> I also see that [MIG] to V13 was done by nobump instead of patch. @rousseldenis

Yes, that's the point.

See : https://pypi.org/project/odoo13-addon-crm-lead-code/#history

No release version !

So, it's not annoying to have a X.X.X.X.1 version but having a release on pypi is better.

pedrobaeza:
@rousseldenis as the version is already bumped manually, why this is not packaged? The problem then is on the packager tool. I don't like to have the first version of a module being 14.0.1.0.1

rousseldenis:
> @rousseldenis as the version is already bumped manually, why this is not packaged? The problem then is on the packager tool. I don't like to have the first version of a module being 14.0.1.0.1

I know. Stephane already explained the limitation.

That's subjective. Some would like to have a released package, some would like a round number. If you have a lead to solve this, everyone could be happy :smiley: 

I explain the process for information to new mergers. To have all the keys in hand

pedrobaeza:
OK, but you know my opinion about all the pip thing, so you know my choice.

HaraldPanten:
Thanks for the info @rousseldenis , it's good to know all points of view and the reasons why of some decisions.

I wouldn't merge very often, as I'm the "newcomer". I will do that just for not blocking "simple modules". Even though, it would be nice to have a general "agreement" for all members.

WDYT?

PR #373, opened 2020-12-03
[12.0][ADD] crm_secondary_salesperson

marcelsavegnago:


crm_secondary_salesperson/views/crm_lead.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2020 - TODAY, Marcel Savegnago <marcel.savegnago@escodoo.com.br> - Escodoo
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record model="ir.ui.view" id="crm_lead_form_view">
+        <field name="name">crm.lead.form (in crm_secondary_salesperson)</field>
+        <field name="model">crm.lead</field>
+        <field name="inherit_id" ref="crm.crm_case_form_view_oppor"/>
```
skeller1, line 10: In V12 there's also a `crm_case_form_view_leads` form view for leads. Maybe we should also add the second salesman field into this view.

crm_secondary_salesperson/models/crm_lead.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class CrmLead(models.Model):
+
+    _inherit = 'crm.lead'
+
+    secondary_user_id = fields.Many2one(
```
skeller1, line 11: Maybe you could add a constraint: user_id <> sesconday_user_id ??????

skeller1 (changes requested):
code and functionally tested locally and on runbot: :+1: 

Left some comments.

Thank you for contributing

bealdav:
Thanks a lot for this contribution. I think it's useful.

Using this module, what are the ways to make kpi, like how many times this salesman won/loose a lead ?

It's not a demand of me just a question, have you considered to use a m2m instead of multiple m2o to make report easier ?

Thanks

skeller1:
Good questions :+1: 
IMO this depends on the company structure (sales workflow).
Different company, different information demand, other KPI definitions.

Maybe  @marcelsavegnago could tell us something about the company that need this additional field. 

In the end reporting should be adapted to the company requirements.

Won/lose ratio could be compared with the linked sales orders (with entered sales men).
So there should exist correct commission rules.

BTW: second sales person should not be required (sql constraint).

crm_secondary_salesperson/models/crm_lead.py
marcelsavegnago: Done

marcelsavegnago:
@bealdav e @skeller1 Great questions :-D

In this case, the module was developed with the sole purpose of linking a second salesperson responsible for the opportunity. However, the possibility of adding features such as the calculation of commissions and KPIs of the second supplier is very positive, but I believe that it can stay on the module's roadmap or even be implemented by other modules.

crm_secondary_salesperson/views/crm_lead.xml
marcelsavegnago: Done!

skeller1 (approved):

bealdav (approved):

marcelsavegnago:
> This PR has the `approved` label and has been created more than 5 days ago. It should therefore be ready to merge by a maintainer (or a PSC member if the concerned addon has no declared maintainer). 🤖

ping

rousseldenis (approved):

PR #374, opened 2020-12-10
[11.0][MIG][WIP] crm_mailchimp: Migration to 11.0

regispirard:


regispirard:
To do : 

1. rewrite def get_default_values in models/mailchimp_settings.py : 

    def get_default_values(self, fields):
        get_param = self.env['ir.config_parameter'].get_param
        return {
            key: get_param('crm_mailchimp.%s' % key)
            for key in ('username', 'apikey')
        }

2. fix tests

PR #376, opened 2020-12-22
[14.0] [ADD] crm_lead_type: Added new module for CRM lead type.

### LINKED ISSUE #369: [RFC] crm_lead_type: Lead Type, for product type/line of business analysis
# Description

Lead Types allow to segment lead by product types or lines of business. It provides a second analysis dimension in addition to the Sales Team.

This is the same concept also available for Sales Orders, with [`sale_order_type`](https://github.com/OCA/sale-workflow/tree/13.0/sale_order_type) 

# Configuration

The Sales Manager can configure the available Lead Types in the CRM App > Configuration > Lead Type menu.

# Usage

On the Lead form, set the Lead Type field with the appropriate product type, selected from the configured list.

On the Lead list and pivot views, select Group By "Type" to group the data by type.

Chandresh-SerpentCS:
Closes #369


Chandresh-SerpentCS:
CC @dreispt

crm_lead_type/models/crm_lead_type.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class CRMLeadType(models.Model):
+    _name = "crm.lead.type"
+    _description = "CRM Lead Type"
+
+    name = fields.Char("Name")
```
skeller1, line 11: Maybe we should make the field translatable like in sale_order_type addon ([here](https://github.com/OCA/sale-workflow/blob/13.0/sale_order_type/models/sale_order_type.py#L25))

crm_lead_type/models/crm_lead.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (C) 2020 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class CrmLead(models.Model):
+    _inherit = "crm.lead"
+
+    lead_type_id = fields.Many2one("crm.lead.type", string="Lead Type")
```
skeller1, line 10: Maybe adding ondelete="restrict"?

Do you want to implement default lead_type logic like in sale_order_type addon (store default lead_type at res.partner)?

crm_lead_type/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
```
skeller1, line 1: We need also sales man read permissions (select crm lead type in crm lead record). Have a look at sale_order_type ir.model.access.csv file:

https://github.com/OCA/sale-workflow/blob/13.0/sale_order_type/security/ir.model.access.csv

skeller1 (changes requested):

crm_lead_type/models/crm_lead_type.py
dreispt: Agree

crm_lead_type/models/crm_lead.py
dreispt: Restrict sounds good.
Adding default logic is an interesting idea, but would require wither a glue module or `some changes to `sale_order_type.
I would leave that our for know until I'm sure the customer using this needs that.

skeller1: Okay, no problem, thank you for your feedback.
Also noticed that the crm_lead_type field is only visible for type = lead, not for type = opportunity (https://github.com/odoo/odoo/blob/15269bcfa8e4d914b2834334b6ff797c5538a400/addons/crm/views/crm_lead_views.xml#L222)

Is this the correct behavior?

crm_lead_type/security/ir.model.access.csv
skeller1: Bildschirmfoto_2020-12-30_13-22-10

dreispt:
@skeller1 @Chandresh-SerpentCS I'm thinking that this module would be best having a dependency with `sale_order_type`.
I dislike having a CRM module force installing the Sales module, but it could be a reasonable first approach here.
Right now I don't want to go into the `sale_order_type` that would be needed to avoid this.

Thoughts?

skeller1:
Maybe I don't understand: Why a dependency with sale_order_type addon? My comments only intent to see how it was programmed in this addon to get the same functionality.
Thank you

dreispt:
To avoid duplicating code.
For example, we shouldn't even be duplicating the Type model, we want to use the same model as `sale_order_type`, because when a Lead is converted to Quotation, we want it to be carried over.

crm_lead_type/models/crm_lead_type.py
dreispt: Actually, we want to use here `sale.order.type`, rather than creating a new Model, so that this field can be passed on to the Quotation.

skeller1:
Okay, but lead types and sale order types could be different in a company. But understand what you want to achieve.

dreispt:
This is intended to extend order types, starting with leads. So having different CRM type goes against the intent.

@Chandresh-SerpentCS We need to rework this. Please add the `sale_order_type` dependency and leverage the Types model added there, and even the Partner default type.
Make it `auto_install: True` please. This properly communicates that this is a glue module between CRM and Sale Order Types.

cvinh:
@dreispt how did it end finally ? we would need the a crm_lead_type as well but not linked to sale_order_type

would it be fancy to have an abstract type model that could be herited in sale_order_type and/or in crm_lead_type ?

PR #380, opened 2021-02-05
[12.0][FIX] crm_claim: _description fields are not translated

ypapouin:
`model_ref_id` dropdown content is not properly translated

This fix add a custom translate function since we need to get `model._description` translation but the default gettext `_` alias only search for `code` and `sql_constraint` translations

Before:
image

After:
image


rousseldenis (approved):
That makes sense

marcelsavegnago (approved):

bodedra (approved):
LG :+1:

HaraldPanten (approved):

PR #385, opened 2021-03-09
[13.0][MIG] crm_secondary_salesperson: Migration to 13.0

victoralmau:
Migration to 13.0 + Add tests

Please @joao-p-marques and @CarlosRoca13 can you review it?

@Tecnativa TT28669


crm_secondary_salesperson/tests/test_crm_lead.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def test_crm_with_distinct_secondary_user_id(self):
+        lead = Form(self.env["crm.lead"])
+        lead.name = "Test"
+        lead.user_id = self.user_admin
+        lead.secondary_user_id = self.user_demo
+        lead = lead.save()
+        self.assertEqual(lead.secondary_user_id, self.user_demo)
+
+    def test_crm_with_secondary_user_id(self):
```
joao-p-marques: What is the difference between this test and the previous one?

joao-p-marques (commented):
Just a small doubt with the tests, but overall everything seems OK with the code and functionally :+1:

crm_secondary_salesperson/tests/test_crm_lead.py
victoralmau: Tests check 3 examples:
- Lead with user_id A + secondary_user_id B
- Lead without user_id + with secondary_user_id B
- Lead with user_id A + secondary_user_id A

joao-p-marques: Yes, but when you don't assign a `user_id`, one is automatically assigned from context. In this case, you end up with `user_id` being the System User being used by the tests. 
In that case, does it really make sense to keep the 2 separate situations?

joao-p-marques (approved):

marcelsavegnago (approved):

CarlosRoca13 (approved):

rousseldenis (approved):
Code review

HaraldPanten (approved):

PR #390, opened 2021-05-07
[13.0][ADD] crm_security_group: Add new category about CRM and create 3 groups to separate Sales and CRM apps

victoralmau:
Add new category about CRM and create 3 groups (Own Documents Only, All Documents and Administrator) to separate Sales and CRM apps

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT29634


crm_security_group/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Crm Security Group",
```
pedrobaeza:
```suggestion
    "name": "CRM Only Security Groups",
```

crm_security_group/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module add new group in Sales to show only CRM (and exclude all about sale orders).
```
pedrobaeza: s/add/adds, but the specifications are not this one, but a totally new module category called CRM for having 3 security groups twins to the Sales one, as we don't want to lose the possibility of restricting the visibility of the documents.

Then, you need to recreate all the ACLs, record rules, and add menu groups according these 3 new security groups.

crm_security_group/views/menu_items.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="crm.crm_menu_root" model="ir.ui.menu">
+        <field
+            name="groups_id"
+            eval="[(4, ref('crm_security_group.group_crm_only'))]"
```
pedrobaeza: This file is not correct according specifications. Please check the other comment.

crm_security_group/security/security.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="group_crm_only" model="res.groups">
+        <field name="name">CRM only</field>
+        <field name="category_id" ref="base.module_category_sales_sales" />
```
pedrobaeza: See above about the new category and groups.

pedrobaeza (changes requested):

victoralmau:
Thanks @pedrobaeza , i have applied some changes to split **Crm** from **Sale**.

crm_security_group/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module add new category about CRM and create 3 groups (Own Documents Only, All Documents and Administrator) to separate Sales and CRM apps
```
joao-p-marques: IMHO this is a little bit more clear:
```suggestion
This module adds new security options to CRM by creating 3 groups (Own Documents Only, All Documents and Administrator) to separate Sales and CRM apps permissions
```

crm_security_group/security/security.xml
```diff
@@ -0,0 +1,73 @@
...
+            eval="[(4,ref('crm_security_group.group_crm_manager'))]"
+            name="groups_id"
+        />
+    </record>
+    <record model="res.users" id="base.user_admin">
+        <field
+            eval="[(4,ref('crm_security_group.group_crm_manager'))]"
+            name="groups_id"
+        />
+    </record>
```
joao-p-marques: Isn't this a bit redundant? As you are already assigning users above to each group...

joao-p-marques (changes requested):
Code and functional review.

A few remarks:
* I can access sales teams just with the CRM permissions (without Sales permissions). I guess that is intended, but just wanted to confirm.
* When without Sales permissions and with `CRM - Own documents only` or `CRM - All documents` I get an "ugly" error when accessing the settings menu: 
```
Sorry, you are not allowed to access documents of type 'Fields that can be used for predictive lead scoring computation' (crm.lead.scoring.frequency.field). This operation is allowed for the groups:
	- CRM/Administrator
	- Sales/User: Own Documents Only - (Operation: read, User: 2)
```
* When I have administration permissions on Sale and none on CRM, I can't access anything CRM related. I thought the idea was to keep the permissions already present from Sale to CRM the same.

There might be some more errors, but this is all I could test so far

The rest seems good though :+1:

pedrobaeza:
Settings shouldn't be visible to others users except "CRM managers".

joao-p-marques:
> Settings shouldn't be visible to others users except "CRM managers"

Yes, that is what I thought, I meant accessing the general settings menu

victoralmau:
Thanks @joao-p-marques , changes done.

joao-p-marques (changes requested):
> * When without Sales permissions and with CRM - Own documents only or CRM - All documents I get an "ugly" error when accessing the settings menu:
> * When I have administration permissions on Sale and none on CRM, I can't access anything CRM related. I thought the idea was to keep the permissions already present from Sale to CRM the same

These 2 still happen :confused:

victoralmau:
> > * When without Sales permissions and with CRM - Own documents only or CRM - All documents I get an "ugly" error when accessing the settings menu:
> > * When I have administration permissions on Sale and none on CRM, I can't access anything CRM related. I thought the idea was to keep the permissions already present from Sale to CRM the same
> 
> These 2 still happen confused

Sorry @joao-p-marques , changes done.

crm_security_group/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "name": "CRM Only Security Groups",
+    "summary": "Add new group in Sales to show only CRM",
+    "version": "13.0.1.0.0",
+    "category": "Customer Relationship Management",
+    "website": "https://github.com/OCA/crm/",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["crm", "sale_crm", "sale_management"],
```
pedrobaeza: Why do you need to depend on `sale_crm` and `sale_management`? That shouldn't be correct.

crm_security_group/views/menu_items.xml
```diff
@@ -0,0 +1,264 @@
...
+            eval="[(6, 0, [ref('sales_team.group_sale_salesman_all_leads')])]"
+        />
+    </record>
+    <record id="sale.view_order_product_graph" model="ir.ui.view">
+        <field
+            name="groups_id"
+            eval="[(6, 0, [ref('sales_team.group_sale_manager')])]"
+        />
+    </record>
+    <record id="sale.view_order_product_pivot" model="ir.ui.view">
```
pedrobaeza: I don't know why do you need this (and the other `sale.` things).

pedrobaeza: I still see this.

victoralmau:
@pedrobaeza Changes done.

pedrobaeza (approved):

rousseldenis (approved):

crm_security_group/security/security.xml
```diff
@@ -0,0 +1,61 @@
...
+    <record id="module_category_crm_crm" model="ir.module.category">
+        <field name="name">CRM</field>
+        <field name="parent_id" ref="crm_security_group.module_category_crm" />
+    </record>
+    <record id="group_crm_own_leads" model="res.groups">
+        <field name="name">User: Own Documents Only</field>
+        <field name="category_id" ref="crm_security_group.module_category_crm_crm" />
+        <field
+            name="implied_ids"
+            eval="[(4, ref('base.group_user')), (4, ref('crm.group_use_lead'))]"
```
pedrobaeza: Second implied group is not correct. That's an option in Odoo set by settings:
```suggestion
            eval="[(4, ref('base.group_user')))]"
```

crm_security_group/views/menu_items.xml
```diff
@@ -0,0 +1,44 @@
...
+    </record>
+    <record id="crm.crm_menu_config" model="ir.ui.menu">
+        <field
+            name="groups_id"
+            eval="[(4, ref('crm_security_group.group_crm_manager'))]"
+        />
+    </record>
+    <!-- It's necessary to show sale menu only if user have sale group!-->
+    <record id="sale.sale_menu_root" model="ir.ui.menu">
+        <field name="groups_id" eval="[(4, ref('sales_team.group_sale_salesman'))]" />
```
pedrobaeza: Not sure this one is really needed, as all the submenus already have a security group, and thus if no visible children, this one is not shown AFAIK.

pedrobaeza (changes requested):
See my comments

victoralmau:
Thanks @pedrobaeza , changes done.

pedrobaeza (approved):

PR #403, opened 2021-10-08
[14.0][MIG] crm_project: Migration to 14.0

victoralmau:
Migration to 14.0

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT30261


crm_project/wizard/crm_lead_convert2task.py
```diff
@@ -19,23 +18,26 @@ def default_get(self, fields):
...
+        comodel_name="crm.lead", string="Lead", domain=[("type", "=", "lead")]
+    )
+    project_id = fields.Many2one(comodel_name="project.project", string="Project")
 
     def action_lead_to_project_task(self):
         self.ensure_one()
         # get the lead to transform
         lead = self.lead_id
-        partner_id = self._find_matching_partner()
+        partner_id = lead._find_matching_partner()
```
pedrobaeza: It should be `partner`, isn't it?

crm_project/wizard/crm_lead_convert2task.py
```diff
@@ -8,7 +8,6 @@ class CrmLeadConvert2Task(models.TransientModel):
     """ wizard to convert a Lead into a Project task and move the Mail Thread """
 
     _name = "crm.lead.convert2task"
-    _inherit = "crm.partner.binding"
```
pedrobaeza: Are we losing partner creation on the fly in this case?

crm_project/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_crm_lead_convert2task,access_crm_lead_convert2task,model_crm_lead_convert2task,project.group_project_user,1,1,1,1
```
pedrobaeza: The one that should have permission is the sales users, not project ones.

crm_project/__manifest__.py
```diff
@@ -4,13 +4,17 @@
 {
     "name": "Lead to Task",
     "summary": "Create Tasks from Leads/Opportunities",
-    "sequence": "19",
```
pedrobaeza: Why removing this sequence? I would leave it, as it should have any meaning (like the module loading order). It's a pity to not have any comment about the why.

pedrobaeza (changes requested):
It seems this wizard is only accessible from leads view, but it would be very interesting to have it also on opportunities view.

crm_project/wizard/crm_lead_convert2task.py
victoralmau: This model has been removed in v14.
`_find_matching_partner()` function (which is now in `crm.lead`), `_partner_assignment()` function also takes care of creating the partner if necessary.

crm_project/__manifest__.py
victoralmau: OK, I keep it then even though we don't know the reason for this need.

victoralmau:
Changes made.
Form (`crm.crm_lead_view_form`) view is the same for opportunities and leads.

pedrobaeza (approved):
The sequence thing comes from the original Odoo module:

https://github.com/odoo/odoo/blame/12.0/addons/crm_project/__manifest__.py#L6

But no clue why

CarlosRoca13 (approved):

PR #405, opened 2021-10-21
[14.0][FIX] Fixes crm lead code : we can define our sequence for multi-company

cvinh:
Before that PR:

- The ir.sequence data is hard coded and we cannot use another sequence
- It's does not support multi company

### DIFF
--- a/crm_lead_code/models/crm_lead.py
+++ b/crm_lead_code/models/crm_lead.py
@@ -20,7 +20,10 @@ class CrmLead(models.Model):
     def create(self, vals_list):
         for vals in vals_list:
             if vals.get("code", "/") == "/":
-                vals["code"] = self.env.ref(
-                    "crm_lead_code.sequence_lead", raise_if_not_found=False
-                ).next_by_id()
-        return super().create(vals_list)
+                seq_date = fields.Date.context_today(self)
+                vals["code"] = (
+                    self.env["ir.sequence"]
+                    .with_company(self.company_id)
+                    .next_by_code("crm.lead", sequence_date=seq_date)
+                )
+        return super(CrmLead, self).create(vals_list)


santostelmo (approved):

santostelmo:
@cvinh can you please check https://github.com/OCA/crm/pull/405

SilvioC2C (approved):

simahawk:
@pedrobaeza or @oihane good for you?

crm_lead_code/models/crm_lead.py
```diff
@@ -16,11 +18,16 @@ class CrmLead(models.Model):
...
-        for vals in vals_list:
-            if vals.get("code", "/") == "/":
-                vals["code"] = self.env.ref(
-                    "crm_lead_code.sequence_lead", raise_if_not_found=False
-                ).next_by_id()
-        return super().create(vals_list)
+    @api.model
+    def create(self, vals):
+        if "company_id" in vals:
+            self = self.with_company(vals["company_id"])
```
pedrobaeza: This can contaminate the call to super, so it's better to reduce the scope and do it in the affected line (also for readability):
```
vals["code"] = self.env["ir.sequence"].with_company(...).next_by_code(
```

crm_lead_code/models/crm_lead.py
```diff
@@ -16,11 +18,16 @@ class CrmLead(models.Model):
...
+        seq_date = None
+        if vals.get("code", "/") == "/":
+            seq_date = fields.Datetime.context_timestamp(
```
pedrobaeza: Why not using `fields.Date.context_today`?

santostelmo:
@pedrobaeza is possible to merge that?

pedrobaeza (changes requested):

crm_lead_code/models/crm_lead.py
cvinh: ok it's done

crm_lead_code/models/crm_lead.py
cvinh: because it's done like that in sale https://github.com/odoo/odoo/blob/0ef9583b27cdce82b484f302a77b8c5a96aa73f8/addons/sale/models/sale.py#L500
and purchase https://github.com/odoo/odoo/blob/0ef9583b27cdce82b484f302a77b8c5a96aa73f8/addons/purchase/models/purchase.py#L217

pedrobaeza: It's not the same, as they use the value of the order date. First of all, you don't need `fields.Datetime.to_datetime`, as you are already handling datetime object (on contrary than the `vals["date_order"]` thing on the upstream examples, that can be strings). Second, don't `import datetime`, as `fields.Datetime.now()` gives the same. And finally, I think `context_today` gives the same combined result. So there's still room to improve.

cvinh: thanks for your detailed explanation, the code was already like that but you're right, as long as it can be improved easily ;-)

cvinh:
I think that one is ready to merge ? @pedrobaeza

crm_lead_code/models/crm_lead.py
```diff
@@ -16,11 +16,14 @@ class CrmLead(models.Model):
...
 
-    @api.model_create_multi
-    def create(self, vals_list):
-        for vals in vals_list:
-            if vals.get("code", "/") == "/":
-                vals["code"] = self.env.ref(
-                    "crm_lead_code.sequence_lead", raise_if_not_found=False
-                ).next_by_id()
-        return super().create(vals_list)
+    @api.model
```
pedrobaeza: Don't remove the `model_create_multi` improvement

cvinh: thank you @pedrobaeza

cvinh:
@santostelmo I think this one is about to be merged

pedrobaeza:
The signature and the code is incorrect (it's not taking into account multiple `vals` entry > `vals_list`)

cvinh:
> The signature and the code is incorrect (it's not taking into account multiple `vals` entry > `vals_list`)

What is the signature please ?

cvinh:
Hello can a psc member apply the 'no stale' label please ?

rafaelbn:
This PR is quite near to merge, is it?

cvinh:
> This PR is quite near to merge, is it?

Only the question about signature from @pedrobaeza

rafaelbn:
@pedrobaeza could you please help and clarify @cvinh what he has to do to get this correct and merged? Thank you! 😄

rafaelbn (approved):
👍

pedrobaeza (approved):

pedrobaeza:
It was changed after my comment, but no reply in the thread, so I was not advertised.

/ocabot merge patch

cvinh:
@pedrobaeza you said incorrect signature i did not fix it because I did not understand

pedrobaeza:
It was that, sorry (the arguments names). Now the problem to merge seems to be other. A template copier update should fix it (in other PR).

cvinh:
Should I rebase my branch ?

pedrobaeza:
Fixed in #480 

/ocabot merge nobump

PR #410, opened 2021-11-26
[13.0][ADD] crm_salesperson_planner

ValentinVinagre:
New Module:
This application allows you to track and schedule salespeople visits to your customers, allowing you to determine which opportunities are going to be dealt with on each visit.


crm_salesperson_planner/data/ir_cron_data.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2021 Sygel - Valentin Vinagre
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+    <record id="ir_cron_create_visits" model="ir.cron">
+        <field name="name">CRM: Create salesperson visits </field>
```
pedrobaeza:
```suggestion
        <field name="name">CRM: Create salesperson visits</field>
```

crm_salesperson_planner/models/crm_salesperson_planner_visit.py
```diff
@@ -0,0 +1,119 @@
...
+            ("done", "Visited"),
+            ("cancel", "Cancelled"),
+            ("incident", "Incident"),
+        ],
+        default="draft",
+    )
+    close_reason_id = fields.Many2one(
+        comodel_name="crm.salesperson.planner.visit.close.reason", string="Close Reason"
+    )
+    close_reason_image = fields.Image(
```
pedrobaeza, line 70: Why not using attachments for this and don't limit to one? Anyways, if using it, please add `attachment=True` for not being stored on the DB.

crm_salesperson_planner/models/crm_lead.py
```diff
@@ -0,0 +1,16 @@
...
+
+class CrmLead(models.Model):
+    _inherit = "crm.lead"
+
+    crm_salesperson_planner_visit_ids = fields.Many2many(
+        comodel_name="crm.salesperson.planner.visit",
+        relation="crm_salesperson_planner_visit_crm_lead_rel",
+        string="Visits",
+        copy=False,
+        domain="[('partner_id', 'child_of', partner_id)]",
```
pedrobaeza, line 15: Isn't this domain going to hide some legit linked visits if changing the partner for whatever reason?

crm_salesperson_planner/models/crm_salesperson_planner_visit.py
```diff
@@ -0,0 +1,119 @@
...
+    clase_reason_notes = fields.Text(string="Close Reason Notes")
+    visit_template_id = fields.Many2one(
+        comodel_name="crm.salesperson.planner.visit.template", string="Visit Template"
+    )
+
+    _sql_constraints = [
+        (
+            "crm_salesperson_planner_visit_name",
+            "UNIQUE (name)",
+            "The name must be unique!",
```
pedrobaeza:
```suggestion
            "The visit number must be unique!",
```

but it's interesting to have another code more? Isn't the unique should come by partner_id and date?

crm_salesperson_planner/models/crm_salesperson_planner_visit.py
```diff
@@ -0,0 +1,119 @@
...
+        ),
+    ]
+
+    @api.model
```
pedrobaeza: Do it `create_multi`

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
```diff
@@ -0,0 +1,397 @@
...
+        readonly=True,
+        copy=False,
+    )
+    partner_id = fields.Many2one(
+        comodel_name="res.partner", string="Customer", required=True,
+    )
+    initial_date = fields.Date(
+        string="First Visit", default=fields.Date.context_today, required=True,
+    )
+    next_visit_date = fields.Date(string="Next Visit Date")
```
pedrobaeza: Shouldn't this be computed?

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
```diff
@@ -0,0 +1,397 @@
...
+    _sql_constraints = [
+        (
+            "crm_salesperson_planner_visit_template_name",
+            "UNIQUE (name)",
+            "The name must be unique!",
+        ),
+    ]
+
+    @api.model
+    def create(self, values):
```
pedrobaeza: `create_multi`

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
```diff
@@ -0,0 +1,397 @@
...
+
+    @api.depends("visit_ids.date")
+    def _compute_last_visit_date(self):
+        for sel in self.filtered(lambda x: x.visit_ids):
+            sel.last_visit_date = sel.visit_ids.sorted(lambda x: x.date)[-1].date
+        self.filtered(lambda x: not x.visit_ids).write({"last_visit_date": False})
+
+    def _compute_visit_ids_count(self):
+        for sel in self:
+            sel.visit_ids_count = len(sel.visit_ids)
```
pedrobaeza: You can win performance, specially on batch computes, with `read_group` call.

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
```diff
@@ -0,0 +1,397 @@
...
+        action["context"] = {
+            "default_partner_id": self.partner_id.id,
+            "default_visit_template_id": self.id,
+            "default_description": self.description,
+        }
+        return action
+
+    # --- Recurrency ---
+
+    def _rrule_serialize(self):
```
pedrobaeza: All of this is a verbatim copy of the `calendar.event` code that is included in Odoo core, and you must reuse it instead of duplicating code that may have unnoticed bugs since the moment you perform the fork. One possible thing is to do inheritance of the model changing the final name.

crm_salesperson_planner/models/res_partner.py
```diff
@@ -0,0 +1,28 @@
...
+    _inherit = "res.partner"
+
+    salesperson_planner_visit_count = fields.Integer(
+        string="Number of Salesperson Visits",
+        compute="_compute_salesperson_planner_visit_count",
+    )
+
+    def _compute_salesperson_planner_visit_count(self):
+        for partner in self:
+            operator = "child_of" if partner.is_company else "="
```
pedrobaeza: Use `read_group` for performance reasons.

crm_salesperson_planner/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,3 @@
+To configure this module, you need to:
+
+* Go to new menu **CRM > Configuration > Salesperson Planner > Close Reasons** and create a minimum of one reason for each type.
```
pedrobaeza: Which types?

crm_salesperson_planner/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This application allows you to track and schedule salespeople visits to your customers, allowing you to determine which opportunities are going to be dealt with on each visit.
```
pedrobaeza:
```suggestion
This application allows you to track and schedule salespeople visits to your customers, allowing you to determine which opportunities are going to be dealt on each visit.
```

crm_salesperson_planner/readme/USAGE.rst
```diff
@@ -0,0 +1,3 @@
+Go to new menu **CRM > Salesperson Planner > My Visits or All Visits** and create a new visit.
+or
+Go to **CRM > Salesperson Planner > Visit Templates** and create a new recurring template for create periodical visits.
```
pedrobaeza: Please expand a bit more this usage section.

crm_salesperson_planner/wizards/crm_salesperson_planner_visit_close_wiz.py
```diff
@@ -0,0 +1,65 @@
...
+
+class CrmSalespersonPlannerVisitCloseWiz(models.TransientModel):
+    _name = "crm.salesperson.planner.visit.close.wiz"
+    _description = "Get Close Reason"
+
+    def _default_new_date(self):
+        visits = self.env["crm.salesperson.planner.visit"].browse(
+            self.env.context.get("active_ids")
+        )
+        return visits.date
```
pedrobaeza, line 15: Possible singleton error? (the same for all)

crm_salesperson_planner/wizards/crm_salesperson_planner_visit_template_create.py
```diff
@@ -0,0 +1,38 @@
...
+from odoo.exceptions import ValidationError
+
+
+class CrmSalespersonPlannerVisitTemplateCreate(models.TransientModel):
+    _name = "crm.salesperson.planner.visit.template.create"
+    _description = "crm salesperson planner visit template create"
+
+    def _default_date_to(self):
+        template = self.env["crm.salesperson.planner.visit.template"].browse(
+            self.env.context.get("active_ids")
```
pedrobaeza: Singleton error

```suggestion
            self.env.context.get("active_id")
```

pedrobaeza (changes requested):
This requires a test suite that covers the main flows.

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
ValentinVinagre: Ups, this field should never have existed.

crm_salesperson_planner/models/crm_salesperson_planner_visit.py
ValentinVinagre: I think one picture is enough. I don't think salespeople take 2-4 pictures, but it's a good point.

crm_salesperson_planner/models/crm_lead.py
ValentinVinagre: I don't think so. I would say when it is a relational field and this has records, even if some of the records don't respect the domain of the relation, it doesn't remove/hide it.

crm_salesperson_planner/models/crm_salesperson_planner_visit.py
ValentinVinagre: If the name and the restriction of this is "client and date", it will be limited to the fact that only one visit per client can be had on the same day and I believe that this should not be limited. For what I see interesting a numbering.

crm_salesperson_planner/models/crm_salesperson_planner_visit.py
pedrobaeza: Then I don't see the need of any numbering. At the end, are you going to say that you have performed the visit number 0568 or the visit to the customer X on the date Y?

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
ValentinVinagre: I thought about doing an inheritance of the model to reuse it, but there are really many things that are not needed and the original functions are not required all, so I was cleaning them and changing them a bit to adapt it to what I need.

crm_salesperson_planner/models/crm_salesperson_planner_visit.py
ValentinVinagre: Said like that it sounds good, but then do we remove the possibility that there may be two visits for the same client on the same day? I see that this possibility may exist.

carlosdauden:
We are working on a similar module, but it is based on calendar.event instead of creating a new data model from scratch, to take advantage of all the "magic" that Odoo has implemented for recurrence.
Why are you copying the methods used by calendar.event instead of inheriting it?

crm_salesperson_planner/models/crm_salesperson_planner_visit.py
pedrobaeza: Yes, but still, you have that as identification, but no need of an extra useless number and no unique should be applied.

crm_salesperson_planner/models/crm_lead.py
pedrobaeza: I'm not sure about that, but if you have investigated it, you tell me.

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
pedrobaeza: No matter if there are things not used. You are not copying code and you only use what you want.

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
```diff
@@ -0,0 +1,406 @@
...
+        [
+            ("1", "First"),
+            ("2", "Second"),
+            ("3", "Third"),
+            ("4", "Fourth"),
+            ("5", "Fifth"),
+            ("-1", "Last"),
+        ],
+        string="By day",
+    )
```
carlosdauden: We are working on a similar module, but it is based on calendar.event instead of creating a new data model from scratch, to take advantage of all the "magic" that Odoo has implemented for recurrence.
Why are you copying the methods used by calendar.event instead of inheriting it?

carlosdauden: Another advantage of using calendar.event directly is that all this information is directly displayed in each salesman's calendar and it can even be automatically synchronized with google calendar

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
carlosdauden: Sorry, I had not seen that it was already being discussed here.
I do not know why these comments do not appear in the conversation tab...

crm_salesperson_planner/models/crm_salesperson_planner_visit.py
ValentinVinagre: Ok i will change it. name = customer+date and I'll change the restriction to be for the name. It's okay?

crm_salesperson_planner/models/crm_lead.py
ValentinVinagre: If some of a partner's contacts change parent, it does not affect the relationship already created between the visit and the opportunity.

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
ValentinVinagre: I continue this conversation in the other one already created.

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
ValentinVinagre: @carlosdauden : this model don't use datetime, only date and priority. The calendar.event use datetime for show the records in the calendar views.
@pedrobaeza : okay.

both: I will look at the way you tell me to use it. Thanks for the comments.

pedrobaeza: I see convenient to order visits inside same day, but if not, you can do it through a calendar.event marking the event as "all day event".

ValentinVinagre: The problem would come when he had more than one visit per day (which is normal). It would fill his entire calendar and it would be difficult to manage if he has to schedule other actions.

pedrobaeza: But that visits are time that you spent, and it should be reflected. Anyways, if they are put as all day, they are folded on the up part.

ValentinVinagre: We have put it together with calendar.event, what do you think?

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
```diff
@@ -0,0 +1,226 @@
...
+    def create(self, vals_list):
+        for vals in vals_list:
+            if vals.get("name", "/") == "/":
+                vals["name"] = self.env["ir.sequence"].next_by_code(
+                    "salesperson.planner.visit.template"
+                )
+        return super().create(vals_list)
+
+    # overwrite
+    def read(self, fields=None, load="_classic_read"):
```
pedrobaeza: Why are you doing all of these overridings? I don't think they are needed, as these are not virtual records. Anyways, always that you make something controversial, document it.

crm_salesperson_planner/models/calendar_event.py
```diff
@@ -0,0 +1,42 @@
...
+class CalendarEvent(models.Model):
+    _inherit = "calendar.event"
+
+    def write(self, values):
+        ret_val = super(CalendarEvent, self).write(values)
+        if ret_val and values.get("start"):
+            for sel in self.filtered(
+                lambda a: a.res_model == "crm.salesperson.planner.visit"
+            ):
+                salesperson_visit_id = self.env["crm.salesperson.planner.visit"].search(
```
pedrobaeza: Do browse over `res_id` instead of the search. Other possibility is to add a one2many here and directly use `mapped`.

Another thing: as the start value is the same for all, please do only one write.

pedrobaeza (changes requested):

crm_salesperson_planner/models/calendar_event.py
ValentinVinagre: Done.

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
ValentinVinagre: Better explained.

pedrobaeza (approved):
Squash commits

ValentinVinagre:
> Squash commits

Done 👍🏻

ValentinVinagre:
Ready to merge :)

PR #411, opened 2021-11-26
[13.0][ADD] crm_salesperson_planner_sale

ValentinVinagre:
This module allows you to manage sales from commercial visits.
Depends https://github.com/OCA/crm/pull/410


crm_salesperson_planner_sales/models/crm_salesperson_planner_visit.py
```diff
@@ -0,0 +1,89 @@
...
+    sale_order_count = fields.Integer(
+        compute="_compute_sale_data", string="Number of Sale Orders", store=True
+    )
+    quotation_count = fields.Integer(
+        compute="_compute_sale_data", string="Number of Quotations", store=True
+    )
+
+    @api.depends("order_ids.state")
+    def _compute_sale_data(self):
+        for sel in self:
```
pedrobaeza: Use `read_group` for performance.

crm_salesperson_planner_sales/models/crm_salesperson_planner_visit.py
```diff
@@ -0,0 +1,89 @@
...
+            sale_order_cnt = 0
+            for order in sel.order_ids:
+                if order.state in ("draft", "sent"):
+                    quotation_cnt += 1
+                if order.state not in ("draft", "sent", "cancel"):
+                    sale_order_cnt += 1
+            sel.quotation_count = quotation_cnt
+            sel.sale_order_count = sale_order_cnt
+
+    def action_sale_quotations_new(self):
```
pedrobaeza: Avoid unneeded plurals in methods

crm_salesperson_planner_sales/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows you to manage sales from commercial visits.
```
pedrobaeza: A bit of usage explanation?

pedrobaeza (changes requested):
Please remove the plural in the module name: `crm_salesperson_planner_sale`. Good point on the gender neutral `salesperson` by the way!

Please complement tests from the other module to be done with an extension for this feature here.

ValentinVinagre:
> 
Hi @pedrobaeza ,
thanks for the review. I think 'salesperson' isn't plural, is the singular word (the plural would be salespeople).

pedrobaeza:
I'm talking about the final `s`, not that word.

ValentinVinagre:
> I'm talking about the final `s`, not that word.

a okay, I will make the change then.

crm_salesperson_planner_sales/models/crm_salesperson_planner_visit.py
ValentinVinagre: I would say that's it, if not ... I could do it with a filtered using the first way.

crm_salesperson_planner_sales/models/crm_salesperson_planner_visit.py
ValentinVinagre: Done

crm_salesperson_planner_sales/readme/DESCRIPTION.rst
ValentinVinagre: I have explained a little more, what do you think?

pedrobaeza:
Please perform all changes and then do the comments at once (you can accumulate then), but this is too hammering. And also we should start by the other module, that is dependency of this one.

PR #420, opened 2022-02-10
[MIG][15.0]crm_lead_vat: Migration V15

bizzappdev:


crm_lead_vat/models/crm_lead.py
```diff
@@ -3,6 +3,11 @@
 
 from odoo import fields, models
 
+from odoo.addons.crm.models.crm_lead import PARTNER_ADDRESS_FIELDS_TO_SYNC
+
+# Add vat in list
+PARTNER_ADDRESS_FIELDS_TO_SYNC += ["vat"]
+
```
yajo: This code is loaded when the module is scanned by Odoo, at boot.

🛑 Thus you shouldn't alter this global variable (actually a pseudo-constant, as stated by the UPPER_CASE name).

crm_lead_vat/models/crm_lead.py
```diff
@@ -17,11 +22,7 @@ def _create_customer(self):
...
 
-    def _prepare_values_from_partner(self, partner):
+    def _prepare_address_values_from_partner(self, partner):
         """Recover VAT from partner if available."""
-        result = super(Lead, self)._prepare_values_from_partner(partner)
-        if not partner:
-            return result
-        if partner.vat:
-            result["vat"] = partner.vat
+        result = super(Lead, self)._prepare_address_values_from_partner(partner)
```
yajo: Following from the above comment, you should undo this diff and leave it as it was before.

This way, we only alter the behavior when a database has the module installed, and not only when it is available.

yajo (changes requested):
@moduon MT-518

rafaelbn:
@bizzappdev are you going to continue with this PR? Thanks! 😄

RikiBalthazar:
def _prepare_address_values_from_partner(self, partner):
        """Recover VAT from partner if available."""
        result = super(Lead, self)._prepare_address_values_from_partner(partner)
        if not partner:
            return result
        if partner.vat:
            result["vat"] = partner.vat
        return result

Hello!
I made this work changing "_prepare_values_from_partner" whit > "_prepare_address_values_from_partner"
I hope this helps

crm_lead_vat/models/crm_lead.py
bizzappdev: changes are done.

PR #424, opened 2022-03-03
[15.0][ADD] crm_won_reason

ajaniszewska-dev:


crm_win_reason/wizard/crm_lead_won.py
```diff
@@ -0,0 +1,17 @@
...
+
+from odoo import fields, models
+
+
+class CrmLeadReason(models.TransientModel):
+    _name = "crm.lead.won"
+    _description = "Get Won Reason"
+
+    win_reason_id = fields.Many2one(
+        "crm.lost.reason", "Won Reason", domain="[('reason_type','=','won')]"
```
ajaniszewska-dev: i don't like the fact that model has still 'lost' in its name even it is used for won reason but ATM functionality is there, maybe can be changed later.

crm_win_reason/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
```
ivantodorovich:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
```
https everywhere :)

anyway there's no need for headers in `__init__.py` files

crm_win_reason/wizard/crm_lead_won.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class CrmLeadReason(models.TransientModel):
+    _name = "crm.lead.won"
+    _description = "Get Won Reason"
+
+    win_reason_id = fields.Many2one(
```
ivantodorovich:
```suggestion
    won_reason_id = fields.Many2one(
```
For consistency with `lost_reason_id`.
Same verb tense

crm_win_reason/models/crm_lost_reason.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class CrmLostReason(models.Model):
+    _inherit = "crm.lost.reason"
+
+    reason_type = fields.Selection(
+        [("won", "Won"), ("lost", "Lost")],
```
ivantodorovich: On the wizard, the domain is hardcoded to either `[('reason_type','=','won')]` or `[('reason_type','=','lost')]`.
So, either:

- Make it required here
- Allow for non-specifcied reason types (global // apply to both won or lost):
    ie: `[('reason_type','in',[False, 'won'])]`

ivantodorovich (approved):
Would you consider adding test coverage?

crm_win_reason/views/crm_views.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" ?>
+<odoo>
+    <record id="crm.crm_lost_reason_action" model="ir.actions.act_window">
+        <field name="name">Win or Lost Reasons</field>
+    </record>
+
+    <!-- Rename menuitem -->
+    <menuitem
```
TDu: If it is just to rename the menu you can probably do 
```
<record id="crm.menu_crm_lost_reason" model="ir.ui.menu">
    <field name="name">Win or Lost Reasons</field>
</record>
```

crm_win_reason/models/crm_lost_reason.py
TDu: Should we have a default value ? I am thinking of already existing records especially (set as `lost`)

TDu (commented):
Feel like adding some tests ?

crm_win_reason/models/crm_lost_reason.py
tarbib: @ajaniszewska-dev good point from @TDu, the previously existing reasons should get "lost" as reason type

crm_win_reason/models/crm_lead.py
```diff
@@ -7,7 +7,7 @@
 class CrmLead(models.Model):
     _inherit = "crm.lead"
 
-    win_reason_id = fields.Many2one("crm.lost.reason")
+    won_reason_id = fields.Many2one("crm.lost.reason")
```
tarbib: @ajaniszewska-dev can you also rename the module and PR to crm_won_reason ?

ajaniszewska-dev: @tarbib  renamed

crm_win_reason/models/crm_lost_reason.py
ajaniszewska-dev: @ivantodorovich @TDu  thank you, comments from review addressed

crm_won_reason/tests/test_crm_lead_reason.py
```diff
@@ -0,0 +1,38 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo.tests.common import TransactionCase
+
+
+class TestCrmLeadReason(TransactionCase):
+    def setUp(self):
```
ivantodorovich: Do this in `setUpClass` instead, so that it runs only once for all tests.

Also add `tracking_disable` to speed them up:

```py
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

crm_won_reason/wizard/crm_lead_won.py
```diff
@@ -0,0 +1,17 @@
...
+    _name = "crm.lead.won"
+    _description = "Get Won Reason"
+
+    won_reason_id = fields.Many2one(
+        "crm.lost.reason", "Won Reason", domain="[('reason_type','in',[False,'won'])]"
+    )
+
+    def action_win_reason_apply(self):
+        leads = self.env["crm.lead"].browse(self.env.context.get("active_ids"))
+        return leads.action_set_won(won_reason_id=self.won_reason_id.id)
```
ivantodorovich: It seems this method isn't tested.
Nitpicking here but you can easily achieve 100% coverage by slightly changing tests

ivantodorovich (approved):

crm_won_reason/wizard/crm_lead_won.py
ajaniszewska-dev: for that probably i'd have to test wizard?

ivantodorovich: that's right

ajaniszewska-dev: @ivantodorovich  I changed but still not close to 100% percent coverage, i need it for customer for today so please approve if code LG to you.

ivantodorovich: Oh, I've already approved on my first review 😅
These are just nitpicking suggestions, IMO it's ready.
Thanks for this module

yankinmax:
@simahawk can we reopen it? @ajaniszewska-dev let get this finally merged, you did a great work, thx

simahawk:
pls rebase

yankinmax (approved):

crm_won_reason/__init__.py
```diff
@@ -1,2 +1,27 @@
...
+
+
+def pre_init_hook(cr):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    # in databases where sale and crm are both installed sale_crm is auto installed
+    # sale_crm declares view crm_case_form_view_oppor which looks for
+    # <xpath expr="//button[@name='action_set_won_rainbowman']"
+    # but crm_won_reason changes the name of action_set_won_rainbowman
+    # so "button cannot be located in parent view" error is raised
+    # crm_won_reason depends only on crm, so we need to remove sale_crm
```
yankinmax: @ivantodorovich this is the issue that happens on this PR. And I thought it can be solve such way. But, nope `sale_crm` is going to be installed after this one.
Do you have any ideas?

crm_won_reason/__init__.py
```diff
@@ -0,0 +1,27 @@
+from . import models
+from . import wizard
+from odoo import api, SUPERUSER_ID
+from logging import getLogger
+
+_logger = getLogger(__name__)
+
+
+def pre_init_hook(cr):
```
simahawk: can you pls move this to `hooks.py` ?

crm_won_reason/__init__.py
simahawk: >  but crm_won_reason changes the name of action_set_won_rainbowman

why?

simahawk (changes requested):

crm_won_reason/__init__.py
yankinmax: I can)

crm_won_reason/__init__.py
yankinmax: because we need to use new action name
```
<xpath
    expr="//button[@name='action_set_won_rainbowman']"
    position="attributes"
>
    <attribute
        name='name'
    >%(crm_won_reason.crm_lead_win_action)d</attribute>
    <attribute name='type'>action</attribute>
</xpath>
```
so name isn't `action_set_won_rainbowman` anymore after `crm_won_reason` is installed

simahawk: Then hide the original one and yours. Isn't this an option?

yankinmax: @simahawk done it and checked on the project, seems good, thanks for the review

crm_won_reason/models/crm_lead.py
```diff
@@ -0,0 +1,17 @@
...
+class CrmLead(models.Model):
+    _inherit = "crm.lead"
+
+    won_reason_id = fields.Many2one("crm.lost.reason")
+
+    # override action to consider won reason
+    def action_set_won(self, **additional_args):
+        res = super().action_set_won()
+        if additional_args:
+            self.write(dict(additional_args))
```
simahawk:
```suggestion
            self.write(additional_args)
```

crm_won_reason/models/crm_lead.py
```diff
@@ -0,0 +1,17 @@
...
+from odoo import fields, models
+
+
+class CrmLead(models.Model):
+    _inherit = "crm.lead"
+
+    won_reason_id = fields.Many2one("crm.lost.reason")
+
+    # override action to consider won reason
+    def action_set_won(self, **additional_args):
```
simahawk: is this override needed at all? There's no logic and all you do is passing additional fields to write. You could do that in your wizard w/o this override. :thinking:

crm_won_reason/wizard/crm_lead_lost.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+from odoo import fields, models
+
+
+class CrmLeadLost(models.TransientModel):
+    _inherit = "crm.lead.lost"
+
+    lost_reason_id = fields.Many2one(
+        "crm.lost.reason", domain="[('reason_type','in',[False,'lost'])]"
```
simahawk:
```suggestion
        "crm.lost.reason", domain="[('reason_type','in',(False,'lost'))]"
```

crm_won_reason/wizard/crm_lead_won.py
```diff
@@ -0,0 +1,17 @@
...
+
+from odoo import fields, models
+
+
+class CrmLeadReason(models.TransientModel):
+    _name = "crm.lead.won"
+    _description = "Get Won Reason"
+
+    won_reason_id = fields.Many2one(
+        "crm.lost.reason", "Won Reason", domain="[('reason_type','in',[False,'won'])]"
```
simahawk:
```suggestion
        "crm.lost.reason", "Won Reason", domain="[('reason_type','in',(False,'won'))]"
```

crm_won_reason/wizard/crm_lead_lost.py
simahawk: you should change only the domain AFAIS

crm_won_reason/wizard/crm_lead_won.py
simahawk:
```suggestion
        res = leads.action_set_won()
        leads.won_reason_id = self.won_reason_id
        return res
```
See msg above

crm_won_reason/models/crm_lead.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class CrmLead(models.Model):
+    _inherit = "crm.lead"
+
+    won_reason_id = fields.Many2one("crm.lost.reason")
```
simahawk, line 10: hmm `won_*` that points to `lost` model... :thinking:

crm_won_reason/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module adds possibility to define lost & won reasons to be applied on leads.
```
simahawk, line 1: the lost reason is already implemented in core AFAIS.

simahawk (commented):
I'm a bit puzzled about some changes but I'm not CRM expert here :)

yankinmax:
@simahawk I should check your suggestions because initially, I was not an author of the module. I've just forced a merge of it. Thanks

yankinmax:
@ajaniszewska-dev this module was developed to have the possibility to submit `won_reason` with the Won action (standard).
From what I see the functionality is similar to `lost_reason`. In the standard we don't have `won_reason`, just "Mark as won" action.

crm_won_reason/models/crm_lead.py
ajaniszewska-dev: i dont remember much after ~1 yr but looking at the card issue was that there was only possibility to define lost reasons, so idea was to have selection field Reason Type (lost or won) so we could assign proper value to any record on lost.reason model (i agree it is stupid because model name is crm.lost.reason...). With this we treated this model as both won&lost reason. Then on lead we would have 2 buttons (won/lost) and wizard where we can choose reason (filtered as lost, won, depending on button used).

cc @yankinmax

crm_won_reason/readme/DESCRIPTION.rst
ajaniszewska-dev: yep, but if i am not mistaken back in time there was no won reason model implemented, that's why we made this model to serve for both lost & won.
If right now it is not the case, then we dont need it.

crm_won_reason/models/crm_lead.py
yankinmax: there is still no separate `won.reason` model, and it's not needed

simahawk (approved):

PR #428, opened 2022-06-08
[15.0][MIG] crm_phonecall_planner: Migration to 15.0 

cesar-tecnativa:
@Tecnativa TT36451


pedrobaeza:
@Tecnativa TT36451

crm_phonecall_planner/wizards/crm_phonecall_planner.py
```diff
@@ -205,7 +205,6 @@ def _schedule_call(self, partner, when):
                 "partner_id": partner.id,
                 "partner_phone": partner.phone,
                 "source_id": self.source_id.id,
-                "tag_ids": [(6, 0, self.tag_ids.ids)],
```
pedrobaeza: Why removing this?

crm_phonecall_planner/wizards/crm_phonecall_planner.py
```diff
@@ -122,7 +122,7 @@ def action_accept(self):
             order="date",
         )
         # Get partners to plan
-        partner_domain = safe_eval(self.res_partner_domain or "[]") + [
+        partner_domain = safe_eval.safe_eval(self.res_partner_domain or "[]") + [
```
pedrobaeza: Previous one wasn't working? This happens usually due to a deployment environment with bad imports.

pedrobaeza (changes requested):
Missing manifest version bump in the migration commit

crm_phonecall_planner/wizards/crm_phonecall_planner.py
cesar-tecnativa: it is not longer used

pedrobaeza (approved):

crm_phonecall_planner/wizards/crm_phonecall_planner.py
chienandalu: What you have to change is the comodel in the field for this wizard as the relation nama has changed from `crm.lead.tag` to `crm.tag`.

crm_phonecall_planner/wizards/crm_phonecall_planner.py
```diff
@@ -0,0 +1,211 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+from odoo.tools.safe_eval import safe_eval
+
+_logger = getLogger(__name__)
+
+
+class CrmPhonecallPlan(models.TransientModel):
+    _name = "crm.phonecall.planner"
```
chienandalu, line 17: The wizard needs an access rules definition. Otherwise, the wizard menu won't be accesible to anyone.

crm_phonecall_planner/wizards/crm_phonecall_planner.py
```diff
@@ -0,0 +1,211 @@
...
+    user_id = fields.Many2one(
+        comodel_name="res.users",
+        string="Responsible",
+    )
+    team_id = fields.Many2one(
+        comodel_name="crm.team",
+        string="Sales Team",
+    )
+    tag_ids = fields.Many2many(
+        comodel_name="crm.lead.tag",
```
chienandalu:
```suggestion
        comodel_name="crm.tag",
```

chienandalu (changes requested):
Thanks @cesar-tecnativa

Some remarks (without them the module isn't usable):

chienandalu (approved):
Tested. Thanks!

PR #429, opened 2022-06-20
[15.0] [MIG] crm_lead_product

xhnilic3:
PR in order to migrate crm_lead_product module to version 15.0


crm_lead_product/tests/test_crm_lead_line.py
```diff
@@ -0,0 +1,168 @@
+# Copyright 2017-19 ForgeFlow S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
+
+from odoo.tests import common
+
+
+@common.at_install(False)
+@common.post_install(True)
```
Rad0van: These 2 prevent the test from being imported and thus ending up with ERROR. Interestingly this isn't caught by the bot so the result is green even it shouldn't be. Check out the result of running test (actually they didn't run). It's something like this:

```
2022-06-20 22:17:04,380 260 ERROR odoo odoo.tests.loader: Can not `import crm_lead_product`. 
  File "/opt/odoo/odoo/tests/loader.py", line 33, in _get_tests_modules
    mod = importlib.import_module('.tests', modpath)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 6[71](https://github.com/OCA/crm/runs/6974767791?check_suite_focus=true#step:8:72), in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 848, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/__w/crm/crm/crm_lead_product/tests/__init__.py", line 3, in <module>
    from . import test_crm_lead_line
  File "/__w/crm/crm/crm_lead_product/tests/test_crm_lead_line.py", line 7, in <module>
    @common.at_install(False)
AttributeError: module 'odoo.tests.common' has no attribute 'at_install'
```

xhnilic3: thank you, I haven't noticed it at all, i'll try to fix that

crm_lead_product/report/crm_product_report.py
```diff
@@ -0,0 +1,104 @@
...
+    name = fields.Char("Lead Name", readonly=True)
+    partner_id = fields.Many2one("res.partner", "Partner/Customer", readonly=True)
+    partner_name = fields.Char("Contact Name", readonly=True)
+    probability = fields.Float(group_operator="avg", readonly=True)
+    stage_id = fields.Many2one("crm.stage", "Stage", readonly=True)
+    team_id = fields.Many2one("crm.team", "Sales Team", readonly=True)
+    type = fields.Char(
+        selection=[("lead", "Lead"), ("opportunity", "Opportunity")],
+        help="Type is used to separate Leads and Opportunities",
+    )
```
ajaniszewska-dev, line 36: `odoo odoo.fields: Field crm.product.report.type: unknown parameter 'selection', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it `

i think there is a mistake with type of the field here?
```suggestion
    type = fields.Selection(
       [("lead", "Lead"), ("opportunity", "Opportunity")],
        help="Type is used to separate Leads and Opportunities",
    )
```

xhnilic3: Hello, thanks for your review, it is possible, i will have a look at it test it and provide feedback. Thank you one more time.

Rad0van: > `odoo odoo.fields: Field crm.product.report.type: unknown parameter 'selection', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it `
> 
> i think there is a mistake with type of the field here?

Funny thing is the same exists in 13.0

frankvogler:
I get an error triggered by the _onchange_lead_line_ids:

_AttributeError: 'crm.lead' object has no attribute 'planned_revenue'_

Reproduce: add a lead line to trigger the onchange.

I suspect `self.planned_revenue` was meant to be `self.expected_revenue`?

emagdalenaC2i:
The field "expected_revenue" is now "prorated_revenue"
and field "planned_revenue" is now "expected_revenue"
See https://github.com/odoo/odoo/commit/8066ab43482a3875d4f432a468aaa017aeea4bba

crm_lead_product/report/crm_product_report.py
```diff
@@ -0,0 +1,104 @@
...
+    date_open = fields.Datetime("Open Date", readonly=True)
+    lost_reason = fields.Many2one("crm.lost.reason", readonly=True)
+    name = fields.Char("Lead Name", readonly=True)
+    partner_id = fields.Many2one("res.partner", "Partner/Customer", readonly=True)
+    partner_name = fields.Char("Contact Name", readonly=True)
+    probability = fields.Float(group_operator="avg", readonly=True)
+    stage_id = fields.Many2one("crm.stage", "Stage", readonly=True)
+    team_id = fields.Many2one("crm.team", "Sales Team", readonly=True)
+    type = fields.Char(
+        selection=[("lead", "Lead"), ("opportunity", "Opportunity")],
```
astirpe, line 34: I just noticed that there's a warning in the log:

```
Field crm.product.report.type: unknown parameter 'selection', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it
 ```
 
It's because the field is actually a Char.

I did not investigate more and I did not make any review of this module. I just accidentally noticed that warning.

astirpe (changes requested):

rousseldenis:
@xhnilic3 Do you mind if we switch to #450 ?

rafaelbn:
Closing without answer in favour of https://github.com/OCA/crm/pull/450

Issue #444, opened 2022-10-10
[14.0] crm_lead_firstname: False positive match when convert lead


Doscaal:
## Module
crm_lead_firstname

## Describe the bug
When you convert a lead in opportunity, there is a look for a potential existing partner relevant.
By default, odoo look on the email and if we don't have a match it look on these criteria partner_name, contact_name, name of the lead.

This is good point but when we use the module crm_lead_firstname, the contact_name field is used to set the firstname. 

The consequence is when we convert the lead, odoo is looking for someone with the same firstname and if there is a match it propose to link it and not to create a new partner.

## To Reproduce
**Affected versions**: 14.0

1. Create partner Paul Taylor
2. Create a lead with Paul as firstname and Graham as lastname.
3. Convert your lead as an opportunity. The system will propose Paul Taylor as a partner to link to the opportunity instead of creating a new partner

**Expected behavior**
IF the complete name match, we propose the contact and if not, we propose to create a new one.

PR #445, opened 2022-10-10
14.0 fix firstname convertion issue

Doscaal:
[#444 ](#444)


rafaelbn:
Not sure about this change technically

The issue is very well explained in https://github.com/OCA/crm/issues/444

What do you think @XanderDeMatteis @oihane ?

PR #446, opened 2022-10-17
[13.0][IMP] crm_project: Copy email_cc from lead to task. 

victoralmau:
Copy email_cc from lead to task.

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT39946

### DIFF
--- a/crm_project/wizard/crm_lead_convert2task.py
+++ b/crm_project/wizard/crm_lead_convert2task.py
@@ -1,5 +1,6 @@
 # Copyright 2010-2020 Odoo S. A.
 # Copyright 2021 Tecnativa - Pedro M. Baeza
+# Copyright 2022 Tecnativa - Víctor Martínez
 # License LGPL-3 - See https://www.gnu.org/licenses/lgpl-3.0.html
 from odoo import api, fields, models
 
@@ -37,6 +38,7 @@ class CrmLeadConvert2Task(models.TransientModel):
             "project_id": self.project_id.id,
             "partner_id": partner_id,
             "user_id": None,
+            "email_cc": lead.email_cc,
         }
         task = self.env["project.task"].create(vals)
         # move the mail thread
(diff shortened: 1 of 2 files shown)


crm_project/wizard/crm_lead_convert2task.py
```diff
@@ -41,6 +42,12 @@ def action_lead_to_project_task(self):
         task = self.env["project.task"].create(vals)
         # move the mail thread
         lead.message_change_thread(task)
+        # add followers from lead
+        for follower in lead.message_follower_ids:
+            task.message_subscribe(
+                partner_ids=[follower.partner_id.id],
+                subtype_ids=follower.subtype_ids.ids,
```
pedrobaeza: Only shared subtypes should be transferred. Not sure if this is done by upper code.

pedrobaeza: Other approach can be to subscribe with default types.

rafaelbn:
IMHO I think we should mix crm_project with followers as:

> Add the followers of the lead in the created task.

It not a general need for all customers.

General is create a task from a lead but not followers.

This force all existing clients with this module to develop another module to prevent it and the task are for project user and leads for sales users.

You are assuming that all sales users have access to projects

Not blocking but... consider it.

pedrobaeza:
The general problem right now is the following:

- A customer writes with someone in CC.
- The CC is reflected as followers in your lead when they are registered in your contacts.
- You convert the lead to task...
- ... and the CC follower is lost!

Indeed, there's a chance with this that the salespersons auto-added to the lead are propagated to the task, but I think that's a little trade-off.

Other option is to get that auto-followers, and substract them from the propagation.

rafaelbn:
You have `Field Name | email_cc` in project_task model, maybe....

pedrobaeza:
Good idea! I see that there's `email_cc` also on `crm.lead`, so now the only question is if this field is filled on the mail gateway.

victoralmau:
Changed the approach to populating the `email_cc` field in task.

pedrobaeza (approved):
Tested also on real instance that the `email_cc` is populated correctly when receiving the mail.

rafaelbn (approved):
/ocabot merge minor

PR #453, opened 2022-11-08
[14.0][FIX] crm_security_group: Fix menu visibility

victoralmau:
FWP from 13.0: https://github.com/OCA/crm/pull/454

Regroup permissions in `Sales` section + Fix menu visibility:
- Show CRM root menu only if user has any CRM permissions.
- Show Reporting menu from CRM only if user has any CRM permissions.

Please @pedrobaeza can you review it?

@Tecnativa TT40356

PR #455, opened 2022-11-09
[15.0][FIX] crm_security_group: Set groups_id in menu items correctly (similar to v14)

victoralmau:
Set `groups_id` in menu items correctly (similar to v14)

Please @pedrobaeza can you review it?

@Tecnativa TT40356

Sorry, I wanted to wait for these changes to be merged in v14 (https://github.com/OCA/crm/pull/453) and then add them to the commit history in v15 migration PR.

### DIFF
--- a/crm_security_group/views/menu_items.xml
+++ b/crm_security_group/views/menu_items.xml
@@ -3,7 +3,7 @@
     <record id="crm.crm_menu_root" model="ir.ui.menu">
         <field
             name="groups_id"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Next time, please drop a note about waiting, but in this case, it was straight-forward that this change should be done directly once it was reviewed in v13. Anyway, let's homogenize it:

/ocabot merge minor

PR #468, opened 2022-12-15
[16.0][MIG] crm_claim: Migration to 16.0

FernandoRomera:


AntoniRomera (approved):

MRomeera (approved):
Ok

FernandoRomera:
@pedrobaeza 
Can you merge this PR, please?
We need by management-system/mgmtsystem_claim.

pedrobaeza (commented):
Why adding `base.group_user` everywhere?

/ocabot migration crm_claim

FernandoRomera:
"Any view with groups_id on it now has to move such groups to the elements of the view. But now you can put groups attribute in a view field instead of isolating it on a view without fear of an access error. Check both things in https://github.com/odoo/odoo/pull/98551 and https://github.com/odoo/odoo/pull/95729."

But maybe I have interpreted it incorrectly.

pedrobaeza:
Yes, but these views don't have any group at view level, so you don't have to change anything. Even more, such group is the base group for backend, so everyone accessing backend has it.

FernandoRomera:
Ok, thanks.

FernandoRomera:
If we dont change anythig, this is error:

<form string="Claim" __validate__="1">
                <header>
                    <field name="stage_id" widget="statusbar" options="{'clickable': 1}"/>

Field 'team_id' used in domain of field 'stage_id' (['|', ('team_ids', '=', team_id), ('case_default', '=', True)]) is restricted to the group(s) base.group_user.

View error context:
{'file': '/__w/crm/crm/crm_claim/views/crm_claim_views.xml',
 'line': 1,
 'name': 'CRM - Claims Form',
 'view': ir.ui.view(558,),
 'view.model': 'crm.claim',
 'view.parent': ir.ui.view(),
 'xmlid': 'crm_case_claims_form_view'}

pedrobaeza:
The error is very weird indeed, but right now I don't know the source of it, so let's continue with it. Is it needed in all the fields you have put it?

FernandoRomera:
In the end, it was only needed in one field. The tests pass without errors.

pedrobaeza:
But you have put it in a lot of fields right now...

ivantodorovich:
Could you remove all the added groups 🙏🏻 ?

FernandoRomera:
@ivantodorovich @pedrobaeza 
Limit to 2 groups.

PR #473, opened 2023-01-23
[15.0][FIX] crm_multicompany_reporting_currency: fix amount

yankinmax:


ivantodorovich (approved):

ajaniszewska-dev (approved):

yankinmax:
@gurneyalex can you pls help to get it merged?

crm_multicompany_reporting_currency/models/crm_lead.py
```diff
@@ -79,5 +79,6 @@ def _compute_amount_multicompany_reporting_currency(self):
             if record.company_currency == record.multicompany_reporting_currency_id:
                 to_amount = record.expected_revenue
             else:
-                to_amount = record.expected_revenue / record.currency_rate
+                inverse_rate = 1.0 / record.currency_rate
+                to_amount = record.expected_revenue / inverse_rate
```
SilvioC2C: Mathematically, to divide by the inverse rate is equivalent to multiplying by the original rate, so this snippet of code is equivalent to:
```suggestion
                to_amount = record.expected_revenue * record.currency_rate
```
which (I believe) is easier to read

SilvioC2C (approved):
Small suggestion, else LGTM

crm_multicompany_reporting_currency/models/crm_lead.py
yankinmax: I'll check @SilvioC2C, thanks

yankinmax:
NOTE: pre-commit is failing because of an error caused by pre-commit initialisation itself. So, it's not related to a module.

simahawk:
https://github.com/OCA/sale-reporting/pull/209#issuecomment-1412177358

And shorten the msg :pray: ?
You can keep the explanation for the description of the commit (that serves as a description for the PR too).

simahawk:
> /ocabot merge nobump

@yankinmax you should use `patch` in this case. Although you cannot merge if you are not a maintainer in the target branch.

/ocabot merge patch

PR #478, opened 2023-02-22
[15.0][FIX] crm_phonecall: Fix error element "schedule_meeting" not found in parent view

sergio-teruel:
cc @Tecnativa 

The parent view has not the element
https://github.com/odoo/odoo/blob/916694786af63ae8d93629041ade73f7d956e0b7/addons/crm/views/res_partner_views.xml#L32

ping @carlosdauden @ernesto-garcia-tecnativa


carlosdauden (approved):

ernesto-garcia-tecnativa (approved):
Code Review OK

PR #489, opened 2023-04-28
[IMP] crm_lead_currency: show customer currency only for opportunities

Josesosa07:
Update fields customer_currency_id  to be shown in the views of both, a lead or an opportunity.
 
image
image

### DIFF
--- a/crm_lead_currency/views/crm_lead_views.xml
+++ b/crm_lead_currency/views/crm_lead_views.xml
@@ -6,14 +6,17 @@
         <field name="model">crm.lead</field>
         <field name="inherit_id" ref="crm.crm_lead_view_form" />
         <field name="arch" type="xml">
-            <field name="partner_id" position="before">
+            <xpath
+                expr="//group[@name='opportunity_partner']//field[@name='partner_id']"
+                position="before"
+            >
                 <field name="customer_currency_id" />
                 <field name="is_same_currency" invisible="1" />
                 <field
                     name="amount_customer_currency"
                     attrs="{'invisible': [('is_same_currency', '=', True)]}"
                 />
-            </field>
+            </xpath>
             <field name="expected_revenue" position="attributes">
                 <attribute name="attrs">
                     {'readonly': [('is_same_currency', '=', False)]}


Josesosa07:
image
image

hugho-ad (approved):

luisg123v (changes requested):
Regarding commit description:
- Title too long and technical
- Missing description

suniagajose:
@Josesosa07 

i suggest this: "show customer currency for both forms leads and opportunities" as commit message

regards,

crm_lead_currency/views/crm_lead_views.xml
```diff
@@ -6,14 +6,28 @@
...
                     name="amount_customer_currency"
                     attrs="{'invisible': [('is_same_currency', '=', True)]}"
                 />
-            </field>
+            </xpath>
+            <xpath
+                expr="//group[@name='opportunity_partner']//field[@name='partner_id']"
+                position="before"
+            >
+                <field name="customer_currency_id" />
```
suniagajose, line 13: maybe readonly here to keep the previous behaviour

what do you think @luisg123v ?

suniagajose (changes requested):

crm_lead_currency/views/crm_lead_views.xml
luisg123v: @suniagajose what do you mean by previous behavior?

If you need it readonly, then it should be set by model so it's applied on every place the field is used

crm_lead_currency/views/crm_lead_views.xml
```diff
@@ -6,14 +6,28 @@
...
+                <field name="is_same_currency" invisible="1" />
```
luisg123v, line 14: I think this is not necessary, because the same field is declared above already

luisg123v (changes requested):

luisg123v:
Hi agree with @suniagajose's suggestion for title.

crm_lead_currency/views/crm_lead_views.xml
suniagajose: @luisg123v 

**current behaviour.**

for leads:
- customer currency is visible and editable

for opportunities:
- customer currency is not visible, therefore is not editable

**my expected behaviour**
for leads:
- customer currency is visible and editable

for opportunities:
- customer currency is  visible, but not editable

by changing this to visible and editable in opportunities we could affect the usual expirience that clients already has by having installed this module from previous version

luisg123v: Then I agree it should be readonly when it's an opportunity. However, that may only be achieved by model, by using the `states` attribute.

suniagajose: @luisg123v @Josesosa07 

sorry, i messed up in my explanation, I meant that in Lead form in 12.0 it is not visible, so in 14.0 it should not be visible either.

12.0
Design Software Info - Odoo


14.0
Design Software Info - Odoo (1)


then, if a customer request that it should be visible in Lead form this should be perfomed in customer app, not here!

crm_lead_currency/views/crm_lead_views.xml
```diff
@@ -6,14 +6,27 @@
         <field name="model">crm.lead</field>
         <field name="inherit_id" ref="crm.crm_lead_view_form" />
         <field name="arch" type="xml">
-            <field name="partner_id" position="before">
+            <xpath
```
suniagajose, line 9: @Josesosa07 

consequently to my comment, this should not be performed here!

crm_lead_currency/views/crm_lead_views.xml
```diff
@@ -6,14 +6,27 @@
...
+            >
                 <field name="customer_currency_id" />
                 <field name="is_same_currency" invisible="1" />
                 <field
                     name="amount_customer_currency"
                     attrs="{'invisible': [('is_same_currency', '=', True)]}"
                 />
-            </field>
+            </xpath>
+            <xpath
```
suniagajose: only this xpath make sense to me

suniagajose (approved):

Josesosa07:
@luisg123v 
The view for only the opportunity was set as v12, and removed for the lead. I think the issues are solved.

suniagajose:
@luisg123v could you approve this???

please

luisg123v (approved):
LGTM :+1:

PR #490, opened 2023-05-08
[16.0][MIG] crm_phonecall: Migration to version 16.0

stefan-tecnativa:
cc @Tecnativa TT42238

@chienandalu @pilarvargas-tecnativa  please review!


crm_phonecall/security/ir.model.access.csv
```diff
@@ -0,0 +1,8 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_crm_phonecall_manager,crm.phonecall.manager,model_crm_phonecall,sales_team.group_sale_manager,1,1,1,1
+access_crm_phonecall_all,crm.phonecall.user,model_crm_phonecall,base.group_user,1,0,0,0
+access_crm_phonecall_sale,crm.phonecall.sale,model_crm_phonecall,sales_team.group_sale_salesman,1,1,1,0
+access_crm_phonecall_report_user,crm.phonecall.report.user,model_crm_phonecall_report,sales_team.group_sale_salesman,1,0,0,0
+access_crm_phonecall_report_manager,crm.phonecall.report,model_crm_phonecall_report,sales_team.group_sale_manager,1,1,1,1
+access_crm_phonecall_partner_manager,crm.phonecall.partner.manager,model_crm_phonecall,base.group_partner_manager,1,1,1,1
+access_crm_phonecall2phonecall,access_crm_phonecall2phonecall,model_crm_phonecall2phonecall,base.group_user,1,0,0,0
```
chienandalu: This permissions aren't enough (maybe a previous version issue). Anyway, the group that can create regular phonecalls, should be able to create phonecalls from this wizard too.

image

crm_phonecall/security/.~lock.ir.model.access.csv#
chienandalu, line 1: This should go away

crm_phonecall/views/crm_phonecall_view.xml
```diff
@@ -235,13 +235,11 @@
...
         <field name="model">crm.phonecall</field>
         <field name="inherit_id" ref="crm_phonecall.crm_case_inbound_phone_tree_view" />
-        <field
-            eval="[(4,ref('crm_phonecall.group_show_form_view'))]"
-            name="groups_id"
-        />
         <field name="arch" type="xml">
             <tree position="attributes">
-                <attribute name="editable" />
+                <t groups="crm_phonecall.group_show_form_view">
```
chienandalu: This doesn't seem to work. Maybe there's a limitation for this technique when overriding attributes.

crm_phonecall/views/res_partner_view.xml
```diff
@@ -17,6 +13,7 @@
                     icon="fa-phone"
                     name="%(crm_phonecall.crm_case_categ_phone_incoming0)d"
                     type="action"
+                    groups="sales_team.group_sale_manager, base.group_partner_manager"
```
chienandalu: I think the space here can be problematic

chienandalu (changes requested):

crm_phonecall/security/ir.model.access.csv
stefan-tecnativa: I was aware of this, but I did not investigate why it had been left like that.

chienandalu: But it's module functionality that doesn't work :thinking: so it's something to be fixed no matter who broke the behavior

crm_phonecall/views/crm_phonecall_view.xml
chienandalu: Still not working. You probably need to override `_get_view` and inject the attribute depending on the user's permission. Some reference on how to override such method (adapt it to your needs):

https://github.com/odoo/odoo/blob/7f6c463ccea5edf35e194e42728f62f4abe5eb9c/odoo/addons/base/models/res_currency.py#L318-L327

crm_phonecall/security/ir.model.access.csv
stefan-tecnativa: Understood, thanks for the advice.

crm_phonecall/views/crm_phonecall_view.xml
stefan-tecnativa: Done, I deleted that record as it is no longer needed with these aproach. Can you check anyway if it is correct please?

CarlosRoca13 (approved):
It LGTM :+1:

crm_phonecall/README.rst
```diff
@@ -0,0 +1,111 @@
...
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *CRM > Phone Calls > Logged Calls > Create*.
```
pilarvargas-tecnativa: *CRM > Phone Calls > Logged Calls > New*

crm_phonecall/README.rst
```diff
@@ -0,0 +1,111 @@
...
+#. If your user has *Show Scheduled Calls Menu* permission, you will see
+   scheduled calls menu too.
+#. In any moment you can schedule another call, schedule a meeting or convert
+   call contact to opportunity.
+#. Calls can be categorized and you can manage categories in *CRM >
+   Configuration > Pipeline > Phone Calls > Categories*.
```
pilarvargas-tecnativa: *CRM >
   Configuration > Phone Calls > Categories*

stefan-tecnativa: Change done. Thanks for the suggestions.

pilarvargas-tecnativa (approved):
thanks :+1:

crm_phonecall/views/crm_phonecall_view.xml
pedrobaeza: You have to move the groups to the elements of the view, instead of the whole view. Does the group still exist?

stefan-tecnativa: The approach of moving the group to the view element didn't work for me, and I haven't found any example where it is done either.

crm_phonecall/models/crm_phonecall.py
```diff
@@ -217,3 +217,16 @@ def action_button_convert2opportunity(self):
         opportunity_id = opportunity.create(self._prepare_opportunity_vals())
         self.write({"opportunity_id": opportunity_id.id, "state": "done"})
         return opportunity_id.redirect_lead_opportunity_view()
+
+    def _get_view(self, view_id=None, view_type="form", **options):
+        """This is done for injecting the attribute depending on the user's permissions
+        as moving the group of the view attribute element didn't work and no similar examples
+        of use where found either"""
```
chienandalu:
```suggestion
        """Inject group check to make the tree editable or not. Apparently it can't be done
        by pure view inheritance and attribute overriding"""

chienandalu (approved):

PR #495, opened 2023-05-22
[16.0][MIG] crm_project

SoporteOperu:


dsolanki-initos (commented):
Only the `sales/Administrator` role is allowed to convert leads to tasks. However, according to the `README.rst` file,'Project User' and 'Salesman' roles have permissions to convert leads to tasks. When converting a lead to a task with the `sales/User: Own Documents` rights, an error occurs indicating that the user does not have the rights to delete the lead. Therefore, could you please update the`README.rst` file to specify the salesperson role and the corresponding rights required for conversion?

rafaelbn:
Hello @SoporteOperu ,

We would like to change the name of this module, many years ago of this development

"Lead to task"
"Convert lead to Task"

crm_project is not the correct name

@pedrobaeza @victoralmau do you agree?

I think about creating in OCA/crm a base_crm_project that creates relation between crm and project so other many OCA modules that need this dependency use it

- crm_timesheet
- crm_lead_to_task <- the name I recommend for this PR this migration "crm_project"
- crm_project_task
- etc...

Opinions are always are really welcome @yajo @EmilioPascual @Shide 

/ocabot migration crm_project

Thank you! 😄

pedrobaeza:
OK to `crm_lead_to_task`.

Rename the whole git history doing:
```
git filter-branch --tree-filter 'if [ -d crm_project ]; then mv crm_project crm_lead_to_task; fi' HEAD
git fetch origin 16.0
git rebase origin/16.0
```

carolinafernandez-tecnativa:
@SoporteOperu could you please rebase?

rafaelbn:
Hello @SoporteOperu are you going to continue this PR? A new update one is in https://github.com/OCA/crm/pull/522 Thanks!

rafaelbn:
Superseed by https://github.com/OCA/crm/pull/522

PR #506, opened 2023-07-13
[ADD] crm_lead_lost_reason

SilvioC2C:
Odoo's `crm` module only allows to use the `crm.lead.lost` wizard for `type = 'opportunity'` records. See https://github.com/odoo/odoo/pull/119979.
This module allows using the wizard for all types of `crm.lead` records.


crm_lead_lost_reason/data/ir_actions_server.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="crm.action_mark_as_lost" model="ir.actions.server">
+        <field name="code">
+if record:
```
gurneyalex:
```suggestion
if records:
```

see https://github.com/odoo/odoo/blob/582ff221acba08c6a602e8c2ca13ad8833972440/addons/crm/data/ir_action_data.xml#L16

gurneyalex (approved):
preapproving, but the fix is needed

Issue #481, opened 2023-03-13
[14+][crm_project_task] New module to generate a project/task from opportunities (crm.lead)


rafaelbn:
Hello,

- We are going to generate a ni OCA module to create or link project from Opportunities.
- The OCA one crm_project don't create or link, just convert (could be named `crm_project_lead_to_task`)
- Opportunity will have 1 new smart-buttons to project and task
- Salesman could create and edit project without having permissions in project app
- Salesman could create a task in an existing project without having permissions in project app

In a separate module
- If project is created then the sales order will be also linked when created with the new quotation button
- When project is the "default" one the sale order will be created without project or analytic.
- Module name: crm_project_task

The scope of the module it to allow salesman to create project and task before sales, pre-sales, commercial staff.
Futhermore, the project once create could be target of more expenses like timesheet, or hr expenses or vendor bill.

### sale_project
https://github.com/odoo/odoo/tree/16.0/addons/sale_project
> This module allows to generate a project/task from sales orders.

### sale_crm
https://github.com/odoo/odoo/blob/16.0/addons/sale_crm

> This module allow to generate a sales order from opportunities

### crm_project 
(We don't want to convert crm.lead to task)
https://github.com/OCA/crm/tree/15.0/crm_project
> This module allows to convert leads/opportunities to tasks.

Best regards,
Rafael

@moduon MT-2363

rafaelbn:
Solved in https://github.com/OCA/crm/pull/507

PR #507, opened 2023-07-18
[15.0][ADD] crm_project_task: Module added to create task from lead/opportunities

EmilioPascual:
- Configure default project in CRM > Configuration > Project.
- In any lead click on the button "Create Task"

Solves https://github.com/OCA/crm/issues/481

MT-3175 @moduon @rafaelbn


crm_project_task/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,4 @@
+To configure this module, you need to:
+
```
rafaelbn, line 2: - Create a Project with right configuration in order that all salesman have access (usualy all employees)

@EmilioPascual see

imagen

crm_project_task/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Emilio Pascual (`Moduon <https://www.moduon.team/>`__)
```
rafaelbn, line 1: Check my issue and all explanations plus testing, etc.. I think I'm also contributor 😄 . Functional and consultants are really needed in OpenSource

https://github.com/readme/featured/open-source-non-code-contributions

rafaelbn (commented):
@EmilioPascual , it is not mandatory that Salesman have access to project, they can create but not see project and all task so, the create should have a sude somewhere 

imagen

crm_project_task/readme/CONFIGURE.rst
EmilioPascual: Done

crm_project_task/readme/CONTRIBUTORS.rst
EmilioPascual: Yes, you are right. Functionals are  essentials.

crm_project_task/models/res_company.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Moduon Team S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    crm_default_project_id = fields.Many2one("project.project")
```
yajo, line 10: Shouldn't this be per crm.team?

cc @rafaelbn

crm_project_task/readme/newsfragments/.gitkeep
yajo, line 1: Remove this file please.

crm_project_task/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_crm_create_task,crm.create.task.user,model_crm_create_task,,1,1,1,1
```
yajo: Add a group. For example, crm salesmen.

crm_project_task/tests/common.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2023 Moduon Team S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
+
+from odoo.tests.common import TransactionCase
+
+
+class CrmProjectTaskCommon(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
```
yajo: This method is small and we don't expect many modules to extend this one, so you can just put the method in `TestCrmProjectTask` directly if you prefer. Not blocking.

crm_project_task/tests/test_crm_project_task.py
```diff
@@ -0,0 +1,30 @@
...
+from odoo.tests import tagged
+
+from .common import CrmProjectTaskCommon
+
+
+@tagged("post_install", "-at_install")
+class TestCrmProjectTask(CrmProjectTaskCommon):
+    def setUp(self):
+        super().setUp()
+
```
yajo:
```suggestion
```

crm_project_task/wizards/crm_create_task.py
```diff
@@ -0,0 +1,54 @@
...
+from odoo import _, fields, models
+from odoo.exceptions import UserError
+
+
+class CrmCreateTAsk(models.TransientModel):
+    _name = "crm.create.task"
+    _description = "Wizard to create task"
+
+    lead_id = fields.Many2one("crm.lead")
+    name_task = fields.Char()
```
yajo: This is more natural in English:

```suggestion
    task_name = fields.Char()
```

crm_project_task/wizards/crm_create_task.py
```diff
@@ -0,0 +1,54 @@
...
+    description = fields.Html()
+
+    def create_task(self):
+        project = self.env.company.crm_default_project_id
+        if not project:
+            raise UserError(
+                _(
+                    "Project not configured in settings,"
```
yajo:
```suggestion
                    "Project not configured in settings, "
```

crm_project_task/tests/test_crm_project_task.py
```diff
@@ -0,0 +1,30 @@
...
+    def test_create_task(self):
+        self.task_name = "Test"
+        self.task_description = "Line1</br>Line2"
+        prev_tasks = self.project.task_ids
+        self.env["crm.create.task"].create(
```
yajo: According to https://github.com/OCA/crm/issues/481:

> Salesman could create a task in an existing project without having permissions in project app

So I think this test would benefit if being executed by a user with only salesman permissions. Then, a green test would be checking that behavior too.

crm_project_task/views/crm_lead.xml
```diff
@@ -0,0 +1,23 @@
...
+        <field name="name">crm.lead.view.form.inherit.crm.project.task</field>
+        <field name="model">crm.lead</field>
+        <field name="inherit_id" ref="crm.crm_lead_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="header" position="inside">
+                <button
+                    string="Create Task"
+                    name="%(wizard_crm_create_task_action)d"
+                    type="action"
+                    class="oe_highlight"
```
yajo, line 15: I don't think this button should be highlighted. It has too much visual weight: 
imagen

Highlighted buttons are meant to lead the user to the next natural step in a workflow. For a lead, the natural workflow is to send a quotation and win the lead. Creating a task is a divergent workflow, not the natural one; so I don't think it should be highlighted.

crm_project_task/wizards/crm_create_task.py
```diff
@@ -0,0 +1,54 @@
...
+            )
+        )
+        self.lead_id.message_post(
+            body=_(
+                "Task <a href=# data-oe-model=project.task "
+                "data-oe-id=%(task)d>%(name)s</a> created.",
+                task=task,
+                name=task.display_name,
+            )
+        )
```
yajo, line 44: When you finish, you should return a client action that directs the user to the created task directly. They should be able to still go back to the lead using breadcrumbs if they want.

crm_project_task/wizards/crm_create_task.py
```diff
@@ -0,0 +1,54 @@
...
+from odoo.exceptions import UserError
+
+
+class CrmCreateTAsk(models.TransientModel):
+    _name = "crm.create.task"
+    _description = "Wizard to create task"
+
+    lead_id = fields.Many2one("crm.lead")
+    name_task = fields.Char()
+    description = fields.Html()
```
yajo, line 15: The default value should be the lead's description:

imagen

yajo (changes requested):
Code & functional review. The module looks quite good, but here you have a bunch of suggestions. Thanks!

crm_project_task/models/res_company.py
EmilioPascual: What do you think about it @rafaelbn?
This could be good improvement

crm_project_task/views/crm_lead.xml
rafaelbn: @yajo for the client that need this module, the creation of the task is previous of the creation of the quotation, so is better option higtligthed. Salesman should create task in the most of the cases.

crm_project_task/models/res_company.py
rafaelbn: > Shouldn't this be per crm.team?

It could be and improvement to add but not to replace I mean:

1. As is is OK, same project for all the teams
2. @yajo proposal could be to improve the module for the case that each crm.team should use a different project. So if project is set in crm.team, this override the general one.

I agree "2" could be an improvement, not requested per any client. So in this moment my opinion is to add this improvement @EmilioPascual to the section ROADMAP and not make it until we have a real request.

I'm going to add @yajo (I didn't want to open a big melon 😄 )

- In OCA many modules need to link o create crm.lead with project, each module add the relation.. example: 

https://github.com/OCA/timesheet/blob/15.0/crm_timesheet/models/crm_lead.py#L13

- The best OCA way should be have a base_crm_project module to make this relation in common to other modules like crm_project_task, crm_timesheet, etc...

- And improvement for this module o other, soon, will be the availability to create a new project from crm.lead

- What happed if we want to use both modules separated?

Don't miss: https://github.com/OCA/crm/issues/481

Conclusion: as you see we can think and spend here many hours so in my opinion maintain as is and let see for Odoo 16 or 17 improvements

crm_project_task/wizards/crm_create_task.py
rafaelbn: @yajo the description of the task don't have and relation with the internal notes of the lead, explanation:

- Internal notes of the lead are comercial notes, only for salesman of the lead, shouldn't be propagated

- Description of the task is what a "project uset" should do/make and nothing about the sales

So this is in fact, intended 👍🏼 

Thank you!

crm_project_task/wizards/crm_create_task.py
rafaelbn: @yajo , the task is not for the salesman, and in fact the salesman don't have access to project app

crm_project_task/readme/newsfragments/.gitkeep
EmilioPascual: Done

crm_project_task/security/ir.model.access.csv
EmilioPascual: Done

crm_project_task/tests/test_crm_project_task.py
EmilioPascual: Done

crm_project_task/wizards/crm_create_task.py
EmilioPascual: Yes, you are right

crm_project_task/tests/test_crm_project_task.py
EmilioPascual: Again you're right

crm_project_task/wizards/crm_create_task.py
EmilioPascual: Again, :)

crm_project_task/views/crm_lead.xml
EmilioPascual: @rafaelbn is right

crm_project_task/wizards/crm_create_task.py
EmilioPascual: It's right @rafaelbn, lead description must be different from task description

crm_project_task/tests/test_crm_project_task.py
```diff
@@ -0,0 +1,80 @@
...
+            {
+                "name": "Test Project",
+                "description": "Test Description",
+            }
+        )
+
+    def test_create_task(self):
+        self.company.crm_default_project_id = self.project
+        self.task_name = "Task Test"
+        self.task_description = "Line1</br>Line2"
```
yajo: As a rule of thumb, we should avoid creating new attributes on `self` in unit tests.

The reason is that, if we forget about them, they will be kept for next tests. This can introduce unwanted side effects.

So, generally, all attributes are created in the setup stage.

```suggestion
        company.crm_default_project_id = self.project
        task_name = "Task Test"
        task_description = "Line1</br>Line2"
```

crm_project_task/tests/test_crm_project_task.py
```diff
@@ -0,0 +1,80 @@
...
+        self.assertEqual(len(prev_tasks) + 1, len(tasks))
+        task = tasks - prev_tasks
+        self.assertEqual(task.name, self.task_name)
+        self.assertEqual(task.project_id, self.company.crm_default_project_id)
+        self.assertEqual(task.partner_id, self.partner)
+
+    def test_create_task_no_project(self):
+        self.company.crm_default_project_id = False
+        self.task_name = "Task Test"
+        self.task_description = "Line1</br>Line2"
```
yajo: Same here.

crm_project_task/models/res_company.py
yajo: > should be have a base_crm_project module to make this relation in common

In this module, instead of creating a relationship with a field, @EmilioPascual logs a message on both sides of the operation (lead and task). This message links one with the other. IMHO it's enough, so we don't need a m2o field.

In `crm_timesheet`, the lead is linked with a project, not with a task AFAICS, so it'd be a different case anyways.

yajo (changes requested):
Just a couple of minor makeups to the test and we're done.

crm_project_task/tests/test_crm_project_task.py
EmilioPascual: Applied.

yajo (approved):
/ocabot merge patch

PR #516, opened 2023-08-16
[16.0][FIX] crm_location: make computes multi-record aware

hildickethan:
Adapt computes to be multi-record aware to avoid singleton errors when creating leads in batch

### DIFF
--- a/crm_location/models/crm_lead.py
+++ b/crm_location/models/crm_lead.py
@@ -22,20 +22,21 @@ class CrmLead(models.Model):
     @api.depends("location_id")
     def _compute_partner_address_values(self):
         res = super()._compute_partner_address_values()
-        if self.location_id:
-            self.update(
+        for lead in self.filtered("location_id"):
+            lead.update(
                 {
-                    "zip": self.location_id.name,
-                    "city": self.location_id.city_id.name,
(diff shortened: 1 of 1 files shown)


crm_location/models/crm_lead.py
```diff
@@ -22,20 +22,22 @@ class CrmLead(models.Model):
...
-            self.update(
-                {
-                    "zip": self.location_id.name,
-                    "city": self.location_id.city_id.name,
-                    "state_id": self.location_id.city_id.state_id,
-                    "country_id": self.location_id.city_id.country_id,
-                }
-            )
+        for lead in self:
+            if lead.location_id:
```
pedrobaeza: And you avoid a big diff:

```suggestion
        for lead in self.filtered("location_id"):
```

pedrobaeza (commented):
And the correct wording would be "multi-record aware compute", not "batch compute", as the computation is not done in a batch way.

hildickethan:
> And the correct wording would be "multi-record aware compute", not "batch compute", as the computation is not done in a batch way.

reworded and added the filtered

pedrobaeza (approved):
Thanks!

ioans73 (approved):
LGTM

PR #520, opened 2023-09-11
[15.0][FIX] crm_phonecall: allow salesman to schedule a call

cubells:
Without this permission a simple salesman can't create scheduled calls because has no permission to create.


pedrobaeza:
You should add a record rule as well for restricting the calls to them.

PR #527, opened 2023-11-13
[16.0][ADD] crm_exception

AungKoKoLin1997:
This module allows you to attach several customizable exceptions to your opportunities.
@qrtl


rinaldifirdaus (approved):
Functional Test: Works as expected.
image

crm_exception/readme/USAGE.rst
yostashiro, line 1: Please use CONFIGURE.md instead.

crm_exception/models/crm_lead.py
```diff
@@ -0,0 +1,36 @@
...
+                    ],
+                ]
+            )
+            return rule_domain
+        return base_rule_domain
+
+    @api.constrains("ignore_exception", "stage_id")
+    def _check_quantity_positive(self):
+        for record in self:
+            record._check_exception()
```
yostashiro: We should actually extend the write() method to ensure that the check is not restricted to certain fields. For example, with the current logic, even if a field is made required for a specific stage by an exception rule, the field value can be removed once the opportunity is saved with that stage.

crm_exception/views/base_exception_views.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="name">exception.rule.form</field>
+        <field name="model">exception.rule</field>
+        <field name="inherit_id" ref="base_exception.view_exception_rule_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='model']" position="after">
+                <field
+                    name="stage_ids"
+                    widget="many2many_tags"
+                    attrs="{'invisible': [('model', '!=', 'crm.lead')]}"
+                    groups="base.group_system"
```
yostashiro: Why assign this group here?

yostashiro (changes requested):

yostashiro:
@AungKoKoLin1997 Please add tests as well.

crm_exception/views/base_exception_views.xml
AungKoKoLin1997: I want to use `model` field in `attrs` and this field can be accessed by `base.group_system`.
https://github.com/OCA/server-tools/blob/97be5cf17cc32f4ef3e21fa493868426a5054b2d/base_exception/views/base_exception_view.xml#L44-L45

yostashiro: @AungKoKoLin1997 We shouldn't need this, then, since the parent node already has the group.

crm_exception/models/crm_lead.py
```diff
@@ -0,0 +1,39 @@
...
+                    base_rule_domain,
+                    [
+                        "|",
+                        ("stage_ids", "in", tuple(self.stage_id.ids)),
+                        ("stage_ids", "=", False),
+                    ],
+                ]
+            )
+            return rule_domain
+        return base_rule_domain
```
yostashiro:
```suggestion
        rule_domain = super()._rule_domain()
        if self.stage_id:
            rule_domain = expression.AND(
                [
                    rule_domain,
                    [
                        "|",
                        ("stage_ids", "in", self.stage_id.ids),
                        ("stage_ids", "=", False),
                    ],
                ]
            )
        return rule_domain
```

Any reason to use tuple here? `tuple(self.stage_id.ids)`

crm_exception/data/crm_exception_data.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="crm_excep_no_partner" model="exception.rule">
```
yostashiro: Shouldn't this be created as demo data?

crm_exception/tests/test_crm_exception.py
```diff
@@ -0,0 +1,52 @@
...
+            self.opportunity.stage_id = self.stage_won.id
+
+        # Test ignore_exception.
+        self.opportunity.ignore_exception = True
+        self.opportunity.stage_id = self.stage_qualified.id
+        self.assertEqual(self.opportunity.stage_id, self.stage_qualified)
+        self.opportunity.stage_id = self.stage_proposition.id
+        self.assertEqual(self.opportunity.stage_id, self.stage_proposition)
+        self.opportunity.stage_id = self.stage_won.id
+        self.assertEqual(self.opportunity.stage_id, self.stage_won)
```
yostashiro:
```suggestion
        self.opportunity.stage_id = self.stage_qualified.id
        self.opportunity.stage_id = self.stage_proposition.id
        self.opportunity.stage_id = self.stage_won.id
```

crm_exception/tests/test_crm_exception.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Check exception only for qualified and won stages
```
yostashiro, line 38: Split the following into a separate test.

yostashiro (changes requested):

yostashiro:
@AungKoKoLin1997 Please add an explanation on the limitation on extending the create method.

crm_exception/models/crm_lead.py
AungKoKoLin1997: If we don't convert to tuple in here, cache lookeup error(`TypeError: unhashable type: 'list'`) will occur from this.
https://github.com/OCA/server-tools/blob/590cf43c30698fc8dd2fca306fe5bb6444ecb903/base_exception/models/exception_rule.py#L81

yostashiro (approved):
Code review.

AungKoKoLin1997:
@oca-crm-maintainers
Is it possible to take a look this PR?

AungKoKoLin1997:
Tests fails are not related with this PR.

rousseldenis (approved):

PR #528, opened 2023-11-13
Update crm_claim.py

yassine-rhimi:
in odoo v 16 we used       @api.model_create_multi
 Not  use api-model in def create 
please merge it


pedrobaeza:
Please use a proper commit message: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

PR #536, opened 2023-11-30
[16.0] [IMP] crm_claim: Fix translate french

yassine-rhimi:
[16.0] [IMP] crm_claim: Fix translate french


pedrobaeza:
Please use Weblate for that: https://odoo-community.org/page/translate

PR #541, opened 2023-12-08
[15.0][FIX] crm_salesperson_planner: Event is removed when the activity linked is removed

sergio-teruel:
Since this commit https://github.com/odoo/odoo/commit/71dc58acfcc4589bc5996b48e157aea6b0f8a609 Odoo remove the event calendar linked to a mail activity. To avoid Odoo remove the calendar event we disassociate the event and activity.

cc @Tecnativa


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #553, opened 2024-03-07
[15.0][FIX] crm_salesperson_planner: Changes to make templates functional

victoralmau:
Changes done:

- [x] Removed the (unnecessary) `recurrence` field from templates.
- [x] Recurrence fields (from `calendar.event`) are set as store=True.
- [x] Create an onchange to avoid inconsistent values by UX.
- [x] Modified the form view to make it more functional (similar to calendar.event).
- [x] Consider in `_get_max_date()` if a deadline (`until`) has been set.

**Before**:
antes

**After**:
despues

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT48238


sergio-teruel (approved):

crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
```diff
@@ -68,14 +69,31 @@ class CrmSalespersonPlannerVisitTemplate(models.Model):
...
-        required=True,
-    )
     last_visit_date = fields.Date(compute="_compute_last_visit_date", store=True)
     final_date = fields.Date(string="Repeat Until")
     allday = fields.Boolean(default=True)
-    recurrency = fields.Boolean(default=True)
+    """Set all compute=_compute_recurrence fields of calendar.event as store=True.
+    We want to manage the value of the fields manually and we don't want to depend
+    on recurrence_id field (only possible with calendar.event).
+    We don't use the recurrency field either because it is unnecessary."""
```
pedrobaeza: This should be comments, not docstring:

```suggestion
    # Set all compute=_compute_recurrence fields of calendar.event as store=True.
    # We want to manage the value of the fields manually and we don't want to depend
    # on recurrence_id field (only possible with calendar.event).
    # We don't use the recurrency field either because it is unnecessary.
```

crm_salesperson_planner/static/description/index.html
```diff
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
pedrobaeza: This is because you locally have an outdated pre-commit?

victoralmau: I have updated pre-commit and this line has been removed. Is this what you meant?

pedrobaeza (approved):
/ocabot merge major

pedrobaeza:
Eh, wait, better not to update pre-commit, as it triggers a lot of errors in PO files.

victoralmau:
> Eh, wait, better not to update pre-commit, as it triggers a lot of errors in PO files.

Un-done change.

PR #557, opened 2024-03-07
[15.0][ADD] crm_won_restrict_per_stage: New module

carolinafernandez-tecnativa:
- Add a field in the opportunity stages settings Show or not "Won" button.
- Modify the Leads form to make the "Won" button visible if the stage in which the lead is located has the option Show or not "Won" button configured.
- Do not drag allowed from the kanban view to the "Won" stage if you do not proceed from a stage with the option checked.

TT48265
@Tecnativa
@pedrobaeza @victoralmau


crm_won_restrict_per_stage/models/crm_lead.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo.exceptions import ValidationError
+
+
+class CrmStage(models.Model):
+    _inherit = "crm.lead"
+
+    show_won_button = fields.Boolean(related="stage_id.show_won_button")
+
+    def write(self, vals):
+        if vals.get("stage_id"):
```
pedrobaeza: `self` can be multi-record

crm_won_restrict_per_stage/models/crm_lead.py
```diff
@@ -0,0 +1,18 @@
...
+            stage = self.env["crm.stage"].browse(vals.get("stage_id"))
+            if stage.is_won and not self.stage_id.show_won_button:
+                raise ValidationError(_("The stage must have checked Show won button"))
```
pedrobaeza: Confusing error message:

```suggestion
                raise ValidationError(_("You can't change to this stage from the current stage."))
```

crm_won_restrict_per_stage/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+    * Carolina Fernandez
```
pedrobaeza, line 3: In general, when someone participates in the design, it's important to put them in the contributors. Decide if I have contributed enough for this module to put me here.

crm_won_restrict_per_stage/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module adds possibility to set CRM stages that are being able to see won button and restrict change to an stage won that is not previous set to see won button
```
pedrobaeza: On RST/MD files, it's interesting to cut at 88 cols the same as in code files.

This can also be rephrased for better understanding and ortography:

```suggestion
This module adds the possibility to set CRM stages that are being able to see the won
button (hiding it for the rest).

It also restricts to change to a won stage won from an stage that doesn't have this
check marked.
```

crm_won_restrict_per_stage/views/crm_views.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" ?>
+<odoo>
+    <record id="crm_stage_form" model="ir.ui.view">
+        <field name="name">crm.stage.form.inherit</field>
+        <field name="model">crm.stage</field>
```
pedrobaeza: Don't mix in the same file views from different models. That's an annoying habit in Odoo basecode that we try to avoid here in OCA.

crm_won_restrict_per_stage/readme/DESCRIPTION.rst
pedrobaeza: There's also missing the file CONFIGURE.rst to explain the menu path where to change the check.

pedrobaeza (changes requested):

carolinafernandez-tecnativa:
@pedrobaeza could you please check it now? Thanks :)

pedrobaeza (approved):

victoralmau (approved):

PR #558, opened 2024-03-08
[15.0][IMP] crm_salesperson_planner: Refactoring of the template to make it usable and with simplified code

victoralmau:
Refactoring of the template to make it usable and with simplified code.

@Tecnativa TT48238

### DIFF
--- a/crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
+++ b/crm_salesperson_planner/models/crm_salesperson_planner_visit_template.py
@@ -8,21 +8,51 @@ from datetime import timedelta
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
+from odoo.addons.base.models.res_partner import _tz_get
+from odoo.addons.calendar.models.calendar_recurrence import (
+    BYDAY_SELECTION,
+    END_TYPE_SELECTION,
+    MONTH_BY_SELECTION,
+    RRULE_TYPE_SELECTION,
+    WEEKDAY_SELECTION,
+)
+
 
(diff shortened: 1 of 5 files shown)


victoralmau:
IMO I think you have to remove the `calendar.event` inheritance in templates. It adds "_nothing_", `write()` method is overwritten, and this https://github.com/odoo/odoo/blob/15.0/addons/calendar/models/calendar_event.py#L489 causes problems. 

What do you think @pedrobaeza ?

pedrobaeza:
I can't say right now, Víctor. I need also steps to reproduce the problem this is fixing, and also a regression test would be advisable.

victoralmau:
> I can't say right now, Víctor. I need also steps to reproduce the problem this is fixing, and also a regression test would be advisable.

Sorry, a small change has been added to see the error https://github.com/OCA/crm/actions/runs/8229098690/job/22499657123?pr=558#step:8:258

Steps to reproduce the error:
- Delete all `calendar.event` records
- Try to create a template.

There is a lot of logic inside `calendar.event` and I think that in the template inheriting from calendar.event doesn't add anything now.

victoralmau:
Ping @pedrobaeza

pedrobaeza:
OK, I see. Yes, I agree on removing the inherit in `crm.salesperson.planner.visit.template`.

victoralmau:
All the changes are already done for everything works as expected, simplified the code and added more tests with specific use cases.

Can you check that the operation is correct? Ping @sergio-teruel

pedrobaeza (approved):

sergio-teruel (approved):

PR #559, opened 2024-03-25
[16.0][IMP] crm_salesperson_planner: Refactoring of the template to make it usable and with simplified code

victoralmau:
FWP from 15.0: https://github.com/OCA/crm/pull/553 + https://github.com/OCA/crm/pull/558

Changes done:

- [x] Removed the (unnecessary) `recurrence` field from templates.
- [x] Recurrence fields (from `calendar.event`) are set as store=True.
- [x] Create an onchange to avoid inconsistent values by UX.
- [x] Modified the form view to make it more functional (similar to calendar.event).
- [x] Consider in `_get_max_date()` if a deadline (`until`) has been set.
- [x] Refactoring of the template to make it usable and with simplified code.

**Before**:
antes

**After**:
despues

Please @sergio-teruel and @pedrobaeza can you review it?

@Tecnativa TT48238


sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge major

PR #563, opened 2024-05-16
[17.0][MIG] crm_klaviyo

astirpe:


pedrobaeza:
You should check https://github.com/OCA/automation/tree/16.0/automation_oca as an alternative to Klaviyo

CasVissers-360ERP:
@pedrobaeza 
We don't like Odoo's e-mail template engine, it's not plug & play enough for end-users. Drag & drop not strong, bad multi-language support.

In addition in a tech stack where all outgoing e-mail are handled by Klaviyo (i.e. Shopify -> Klaviyo) it's preferred to keep all outgoing communication from a central place. Also to ensure equal layouts. In addition we want to make sure that a customer doesn't possibly receive 5 e-mails in 10 minutes.

pedrobaeza:
OK, I was just saying, but I think that going to an heterogeneous platform and waste a lot of time connecting them instead of improving Odoo is not the way to go.

I don't see the rest of the reasons you state, like "possibly receive 5 e-mails in 10 minutes". That's a misuse of Odoo, not an issue of Odoo itself.

CasVissers-360ERP:
>> I don't see the rest of the reasons you state, like "possibly receive 5 e-mails in 10 minutes". That's a misuse of Odoo, not an issue of Odoo itself.

Odoo sending an e-mail, Shopify sending one, Sendcloud sending one, Gorgias sending one. 

>> OK, I was just saying, but I think that going to an heterogeneous platform and waste a lot of time connecting them instead of improving Odoo is not the way to go.

We work with big e-com parties opting for a best of breed stack. We want to use Odoo as a transactional ERP.

pedrobaeza:
Again that's because an heterogeneous system. I'm doing decent e-com implementations with Odoo, but everyone decides their way, and their pros and cons. And please don't do advertising here. Hiding everything as off-topic.

PR #565, opened 2024-05-27
[16.0][FIX] crm_claim: Delete create method from crm.claim

pilarvargas-tecnativa:
It's setting a context when creating records for the default team_id and it's not necessary, it's redundant because the team_id field already has a default that defines it. This code is not necessary.

cc @Tecnativa TT49102

@carolinafernandez-tecnativa @pedrobaeza please review

### DIFF
--- a/crm_claim/models/crm_claim.py
+++ b/crm_claim/models/crm_claim.py
@@ -143,13 +143,6 @@ class CrmClaim(models.Model):
         if self.stage_id:
             self.team_id = self.categ_id.team_id
 
-    @api.model
-    def create(self, values):
-        ctx = self.env.context.copy()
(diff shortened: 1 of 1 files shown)


carolinafernandez-tecnativa (approved):

pedrobaeza (changes requested):
I don't have clear the intention of this code, but the transformation to multi-record is not correct. You are putting one of the teams of one of the records with the context for all of them.

Please dig in the intention of that code, and maybe remove it.

pilarvargas-tecnativa:
> Please dig in the intention of that code, and maybe remove it.

Revised, no intention is found in this code since it is not used elsewhere in the model and also a default team_id is already set in the registers.

pedrobaeza (approved):
/ocabot merge patch

PR #569, opened 2024-06-07
[16.0][FIX] crm_won_restrict_per_stage: Add ValidationError to onchange to prevent display incorrect UX changes

victoralmau:
Add ValidationError to onchange to prevent display incorrect UX changes (stage)

**Before**
antes

**After**
despues

Please @pedrobaeza can you review it?

@Tecnativa TT49558

### DIFF
--- a/crm_won_restrict_per_stage/models/crm_lead.py
+++ b/crm_won_restrict_per_stage/models/crm_lead.py
@@ -1,7 +1,7 @@
 # Copyright 2024 Tecnativa - Carolina Fernandez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
 
-from odoo import _, fields, models
+from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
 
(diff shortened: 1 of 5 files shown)


crm_won_restrict_per_stage/models/crm_lead.py
```diff
@@ -10,6 +10,18 @@ class CrmStage(models.Model):
...
 
+    @api.onchange("stage_id")
+    def _onchange_stage_id(self):
+        for item in self:
+            if (
+                item != item._origin
+                and item.stage_id.is_won
+                and not item._origin.stage_id.show_won_button
+            ):
+                raise ValidationError(
```
pedrobaeza, line 26: A raise inside an onchange is not very advisable. Why the write overriding doesn't serve?

pedrobaeza (commented):
Please take the occasion to put the proper maintainer.

crm_won_restrict_per_stage/models/crm_lead.py
victoralmau: I agree, I don't like a raise in an onchange, but if you look at the mini-video in the description you can see that without this change, even if the error is shown, the stage (and everything related) is changed even if it is not saved with the confusion it causes by UX. 

Any other suggestions?

crm_won_restrict_per_stage/models/crm_lead.py
```diff
@@ -10,6 +10,18 @@ class CrmStage(models.Model):
 
     show_won_button = fields.Boolean(related="stage_id.show_won_button")
 
+    @api.onchange("stage_id")
+    def _onchange_stage_id(self):
+        for item in self:
```
pedrobaeza:
```suggestion
    def _onchange_stage_id(self):
        """Do it this way for avoiding a UI glitch if we let act
        the exception in the write, as the stage is changed in the
        interface, but the message appears and the record is kept
        unsaved.
        """
        for item in self:
```

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #576, opened 2024-06-21
[16.0][ADD] crm_required_loss_reason: new module to make required the loss reason of a lead/opportunity

alan196:
This PR adds a new simple module that make required the loss reason when the lost button raises the loss wizard


alan196:
The test error are not related to this module

PR #577, opened 2024-06-27
[16.0][ADD] crm_create_project: Allow create project from lead or opportunity

EmilioPascual:
This module was developed because sometimes you want to create the project from the crm without creating a sales order, for example when you want to plan a project.

It will be useful for you if you want to create a project before creating a quotation and confirming it or if you want to estimate, plan or know if you have the capacity before making the sale.

@rafaelbn @fcvalgar @yajo @Shide please review if you want. Thanks.

MT-6304 @moduon


EmilioPascual:
Error in tests is not related with this module
```
2024-06-27 14:25:01,687 312 ERROR odoo odoo.addons.crm_salesperson_planner.tests.test_crm_salesperson_planner_visit_template: FAIL: TestCrmSalespersonPlannerVisitTemplate.test_05_repeat_weeks
  File "/__w/crm/crm/crm_salesperson_planner/tests/test_crm_salesperson_planner_visit_template.py", line 215, in test_05_repeat_weeks
    self.assertEqual(self.visit_template_base.state, "done")
AssertionError: 'in-progress' != 'done'
- in-progress
+ done
```

fcvalgar (approved):
Functional review
LGTM, thank you @EmilioPascual 

[OCA | [16.0][ADD] crm_create_project: Allow create project from lead or opportunity - MT-6304 - Watch Video](https://www.loom.com/share/c739edfb683247358743a16025611096)

crm_create_project/wizards/crm_create_project.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+        self.lead_id.project_id = project
+        project.message_post(
+            body=_(
+                "Project created from lead/opportunity "
+                "<a href=# data-oe-model=crm.lead data-oe-id=%(lead)d>%(name)s</a>.",
+                lead=self.lead_id,
+                name=self.lead_id.name,
+            )
+        )
```
Shide: maybe you can create message with origin link
```suggestion
        project.message_post_with_view(
            'mail.message_origin_link', 
            values={'self': project, 'origin' self.lead},
            subtype_id=self.env.ref('mail.mt_note').id, 
            author_id=self.env.user.partner_id.id,
        )
```

crm_create_project/wizards/crm_create_project.py
```diff
@@ -0,0 +1,48 @@
...
+        if self.lead_id:
+            self.lead_id.message_post(
+                body=_(
+                    "Project <a href=# data-oe-model=project.project "
+                    "data-oe-id=%(project)d>%(name)s</a> created.",
+                    project=project,
+                    name=project.display_name,
+                )
+            )
```
Shide: Maybe you can use https://github.com/OCA/server-ux/tree/16.0/mail_message_destiny_link_template

```suggestion
            self.lead_id.message_post_with_view(
                'mail_message_destiny_link_template.message_destiny_link',
                values={'self': self.lead_id, 'destiny': project},
                subtype_id=self.env.ref('mail.mt_note').id,
                author_id=self.env.user.partner_id.id,
            )
```

crm_create_project/wizards/crm_create_project.py
Shide: An also check lead_Id as below

Shide (changes requested):
At least the origin link mail message post

crm_create_project/__manifest__.py
yajo, line 1: issue(blocking): Please rename the module to `crm_project_create`. The custom is to put in front the name of the modules that it extends.

crm_create_project/readme/CONTEXT.md
```diff
@@ -0,0 +1,14 @@
...
+    Audience: end users.
+
+    Purpose: help them evaluate if this module is useful in their context:
+
+    - Why this module exists.
+    - Business requirement that generated the need to develop this module.
+    - Point to other related modules.
+    - Explain how those other modules can complement this one.
+-->
+
```
yajo: issue: remove template garbage please.

```suggestion
```

crm_create_project/wizards/crm_create_project.py
```diff
@@ -0,0 +1,48 @@
...
+    _description = "Wizart to create Project from Lead/Opportunity"
+
+    project_name = fields.Char()
+    project_description = fields.Html()
+    lead_id = fields.Many2one("crm.lead")
+
+    def create_project(self):
+        project = (
+            self.env["project.project"]
+            .sudo()
```
yajo: thought: This sudo puts me to think... The salesman should have permissions to create projects? How useful is this module if you can create a project to plan it, but can't edit it later? Maybe this model should be only available to users with CRM *and* project permissions?

crm_create_project/views/crm_lead.xml
```diff
@@ -0,0 +1,26 @@
...
+                    string="Create Project"
+                    name="%(crm_create_project.crm_create_project_action)d"
+                    type="action"
+                    class="oe_highlight"
+                    context="{'default_project_name': name, 'default_lead_id': id}"
+                    attrs="{'invisible': ['|', ('active','=',False), ('project_id', '!=', False)]}"
+                />
+            </xpath>
+            <xpath expr="//group[@name='Misc']" position="inside">
+                <field name="project_id" readonly="1" />
```
yajo: question: why readonly? What if the project is created by somebody else and you just want to link it here?

crm_create_project/views/crm_lead.xml
```diff
@@ -0,0 +1,26 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="header" position="inside">
+                <button
+                    string="Create Project"
+                    name="%(crm_create_project.crm_create_project_action)d"
+                    type="action"
+                    class="oe_highlight"
+                    context="{'default_project_name': name, 'default_lead_id': id}"
+                    attrs="{'invisible': ['|', ('active','=',False), ('project_id', '!=', False)]}"
+                />
```
yajo: issue: on runbot, there's a permissions issue.

image
image
image

yajo (changes requested):

crm_create_project/wizards/crm_create_project.py
EmilioPascual: This module is designed to enable the salesperson to create projects from lead/opportunity. 
Generally, the sales person will not edit the project created from the opportunity. The project will be used by a person with project permissions, occasionally the sales person may be that user if they have project permissions, but usually it will be another user such as a project manager.

crm_create_project/views/crm_lead.xml
EmilioPascual: This is not a glue module linking the crm_lead and project_project models. The field is for the project created, if it could be updated it would not be the one created from the lead/opportunity.
Maybe it could be interesting to change the name of the field to project created, what do you think?

crm_create_project/views/crm_lead.xml
EmilioPascual: The group shown in the image is not odoo core, it is from the crm_security_group module.
The permission for the wizard is for the group *Sales: User: Own Documents Only* which I think is the minimum group the project should be able to create. It could be for all groups but I think there should be a minimum restriction for the group *Sales: User: Own Documents Only*.
I don't think it's right to make this module depend on crm_security_group.

crm_create_project/__manifest__.py
EmilioPascual: Rename done

crm_create_project/wizards/crm_create_project.py
yajo: OK that makes sense. But in that case, please make it clearer in the module. It should be explained in the readme. Maybe a warning message in the wizard would help too (something like: "you won't be able to edit this project after creating it", displayed for non-project-manager salespeople).

crm_create_project/views/crm_lead.xml
yajo: Oh of course, I didn't realize that.

crm_create_project/views/crm_lead.xml
yajo: I understand the use case, but IMHO it might not reflect always the reality.

For example, imagine a customer wants a very specific project. Salesperson doesn't have permission to create a custom project, so he calls the project manager and asks him to create it. Once done, salesperson comes back to the lead to link that project, but it's not possible.

As long as we cover the use case and the field's purpose is well explained, I see no strong reason to forbid users to use it differently. Odoo isn't in general so strict.

rafaelbn: Please @EmilioPascual add this discussions to ROADMAP o KNOWISSUES please and go ahead @yajo . I guess with use the product evolves, we cannot cover all use cases all the time. The most importat is evolve! OK? 😄 ❤️ What do you think?

EmilioPascual: I was thinking about that, maybe @yajo is right. I'm going to try with no readonly field, it may be good for planning extensions to previously created projects.

crm_create_project/wizards/crm_create_project.py
EmilioPascual: Added warning and updated readme, it's a good issue.

crm_project_create/readme/USAGE.md
```diff
@@ -0,0 +1,11 @@
...
+
+To use this module, you need to:
+
+1. Go to CRM > Sales > My Pipeline
+1. Create a new opportunity or join an existing one.
+1. Click on **Create project** in header.
+1. A wizard is shown to fill project name and/or project description.
+1. Click on **Create project**.
+1. You can see project created in page **Extra Information**, inside group **Misc**.
+1. Is you has not group *Project: Administrator* you can edit project after creating.
```
yajo:
```suggestion
1. Is you have *Project: Administrator* permissions, you can edit the project after creating it.
```

crm_project_create/wizards/crm_create_project.py
```diff
@@ -0,0 +1,56 @@
...
+            "project.group_project_manager"
+        ),
+    )
+
+    @api.depends_context("uid")
+    def _compute_show_warning(self):
+        for wizard in self:
+            wizard.show_warning = not self.env.user.has_group(
+                "project.group_project_manager"
+            )
```
yajo: you can remove all this.

```suggestion
```

crm_project_create/wizards/crm_create_project.xml
```diff
@@ -0,0 +1,37 @@
...
+<!-- Copyright 2024 Moduon Team S.L.
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0) -->
+<data>
+    <record id="crm_create_project_view_form" model="ir.ui.view">
+        <field name="name">CRM Create Project Form</field>
+        <field name="model">crm.create.project</field>
+        <field name="arch" type="xml">
+            <form>
+                <p
+                    attrs="{'invisible': [('show_warning', '=', False)]}"
```
yajo: ... and just add this here:

```suggestion
                    groups="!project.group_project_manager"
```

It will make the element invisible for everyone except project managers.

crm_project_create/wizards/crm_create_project.xml
```diff
@@ -0,0 +1,37 @@
...
+                >You won't be able to edit this project after creating it</p>
+                <group>
+                    <field name="project_name" />
+                    <field name="project_description" type="html" />
+                    <field name="show_warning" invisible="1" />
```
yajo: ... and remove this too

```suggestion
```

yajo (changes requested):

yajo:
FWIW the failure in tests is from https://github.com/OCA/crm/issues/589

yajo (approved):

PR #584, opened 2024-07-19
[17.0][MIG] crm_lead_to_task: Migration to version 17.0

carolinafernandez-tecnativa:
- Standard migration to v17
- Fw port fix https://github.com/OCA/crm/pull/587

@Tecnativa
TT50053
@pilarvargas-tecnativa @victoralmau


crm_lead_to_task/wizard/crm_lead_convert2task.py
```diff
@@ -36,7 +36,6 @@ def action_lead_to_project_task(self):
         vals = {
             "name": lead.name,
             "description": lead.description,
-            "email_from": lead.email_from,
```
pilarvargas-tecnativa: Isn't this field now called `email_cc`?

carolinafernandez-tecnativa: Yes, email_from disappears. email_cc is used above, check https://github.com/OCA/crm/pull/584/commits/9a1f0aaf542b12882bbfc598fd055e656173ce74#diff-e7c057e506efe24fd0e6c307c1c712375b6d262f8919b0bb05851d275a93ba05R19

pilarvargas-tecnativa (approved):

carolinafernandez-tecnativa:
ping @pedrobaeza

carolinafernandez-tecnativa:
@victoralmau @pedrobaeza can you review please?  Thanks :)

crm_lead_to_task/tests/test_crm_project.py
```diff
@@ -1,24 +1,24 @@
...
 
-class TestCrmProject(common.TransactionCase):
+class TestCrmProject(BaseCommon):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
         cls.lead = cls.env["crm.lead"].create(
             {
-                "name": "Test lead",
+                "name": "Test Lead",
```
pedrobaeza: Why changing this? Even if we can say that it's the correct upper-casing (which is not true IMO), you shouldn't increase the diff for changes not bringing value to the migration itself.

crm_lead_to_task/tests/test_crm_project.py
```diff
@@ -1,24 +1,24 @@
...
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
         cls.lead = cls.env["crm.lead"].create(
             {
                 "name": "Test lead",
                 "description": "Description",
-                "email_from": "test@example.org",
-                "partner_name": "Test partner",
+                "partner_name": "Test Partner",
```
pedrobaeza: You are still changing this upper-casing. When I put a comment, it applies to all the occurrences of the same type. It's for not repeating the same comment everywhere.

carolinafernandez-tecnativa: This is necessary because when creating task from lead, it will search for partner related to lead and search has changed from odoo 16.0 to 17.0 please check this https://github.com/odoo/odoo/blob/16.0/addons/crm/models/crm_lead.py#L1797
https://github.com/odoo/odoo/blob/17.0/addons/crm/models/crm_lead.py#L1842

partner return is with upper casing

pedrobaeza: I still don't see the relation, as the partner doesn't exist, so it's created the same way you put on `partner_name`. Can you let it as it was to see the problem in CI?

carolinafernandez-tecnativa: Please check now :)

pedrobaeza: The problem was to use a generic partner name, which collides with another existing one. I have pushed the definitive version which doesn't fail.

pedrobaeza (approved):
/ocabot merge nobump

PR #587, opened 2024-07-22
[15.0][FIX] crm_project: Don't remove old lead, but archive

pedrobaeza:
Plain salesmen can't unlink leads, so let's archive it instead, as it was in previous versions.

@Tecnativa TT50262


carolinafernandez-tecnativa (approved):
LGTM, Thanks :) Please check test error, but change its okay.

victoralmau (approved):

PR #596, opened 2024-09-16
Fix permissions issue when clicking phonecalls in crm.lead

PNurm:


crm_phonecall/models/crm_lead.py
```diff
@@ -25,7 +25,7 @@ def _compute_phonecall_count(self):
 
     def button_open_phonecall(self):
         self.ensure_one()
-        action = self.env.ref("crm_phonecall.crm_case_categ_phone_incoming0")
+        action = self.env.ref("crm_phonecall.crm_case_categ_phone_incoming0").sudo()
```
pedrobaeza: The usual thing to do in these cases is to use `_for_xml_id` method instead of `ir.model.data` for reading in one shot without permissions problem the action.

PNurm: Thank you for correcting. I pushed a commit with the correction applied.

PNurm:
Can this be merged now?

PR #599, opened 2024-09-27
[16.0][FIX] crm_security_group : crm_team full access for crm_manager

quentinDupont:
CRM teams (called Sales Teams) permissions access are not fit as usual, it might be a typo ?

**At the moment** 

CRM base user (User : Own documents Only + User : All documents) can fully modified `crm_team`

**With this PR**

CRM base user (User : Own documents Only + User : All documents) can read  `crm_team`
CRM manager (Administrator) can fully modified  `crm_team`

What do you think @victoralmau ?


legalsylvain (approved):
Agree with that proposal. 
changing configuration is for manager in the Odoo logic.

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #601, opened 2024-10-11
[17.0][IMP] crm_lead_code: space get removed when other view changes name field

CRogos:
I have another module which extends the same view like this: 

```
    <record id="crm_case_kanban_view_leads" model="ir.ui.view">
        <field name="name">crm.show.fields</field>
        <field name="inherit_id" ref="crm.crm_case_kanban_view_leads" />
            <field name="name" position="attributes">
                <attribute name="invisible">1</attribute>
            </field>
            <field name="name" position="after">
                <field name="display_name" />
            </field>
        </field>
    </record>
```

This causes that the space (&amp;nbsp;) after the code field is somehow removed. 
image
image

This change prevents this side effect.

### DIFF
--- a/crm_lead_code/views/crm_lead_view.xml
+++ b/crm_lead_code/views/crm_lead_view.xml
@@ -48,7 +48,7 @@
         <field name="inherit_id" ref="crm.crm_case_kanban_view_leads" />
         <field name="arch" type="xml">
             <field name="name" position="before">
-                <field name="code" />&amp;nbsp;
+                <field name="code" /><span class="p-1" />
             </field>
         </field>
     </record>


MohamedOsman7 (approved):
LGTM

crm_lead_code/views/crm_lead_view.xml
```diff
@@ -48,7 +48,7 @@
         <field name="inherit_id" ref="crm.crm_case_kanban_view_leads" />
         <field name="arch" type="xml">
             <field name="name" position="before">
-                <field name="code" />&amp;nbsp;
+                <field name="code" /><span> </span>
```
flotho: Hi, thanks for this PR.
IMHO  `<span/>` should be better

flotho (changes requested):

crm_lead_code/views/crm_lead_view.xml
CRogos: `<span/>` does not add the extra SPACE.
Or is there a css class which we could apply?

CRogos: Ok found a solution: `<span class="p-1" />` 
thank you for your feedback.

flotho (approved):
Hi,
Thanks, much better!

CRogos:
@pedrobaeza would you merge?

PR #608, opened 2024-11-21
[18.0][MIG] crm_lead_product: Migration 18.0

fredericgrall:


lc-obs:
Hey there,
I just checked on the module in V17, and it would be exactly what we need for a customer on V18. Unfortunately, I can't code myself. Do you happen to have any news on when it might be available? Would really appreciate it. Have a nice day :)

etobella (approved):
Can you squash the two migration commits in a single one?

joelkm (approved):
lgtm

joelkm:
Thanks @etobella !!!

etobella:
@fredericgrall can you squash the commits of migration as suggested?

@joelkm if in a week, the comments have not been attended, you can make a new PR keeping the same history of this one and applying the same changes if you want to speed it up. This is called superseeding, however, the right way to do it is to give some time to the current contributor to make the change.

etobella:
Superseeded by #648

PR #618, opened 2024-12-16
[17.0][IMP] crm_project_create: Implementation to 17.0

Gokac9:
Add option to duplicate an existing project with a new name and description. Useful for creating similar projects with the same tasks.
Video crm_project_create


manuelcalerosolis (approved):
Functional review approve

EmilioPascual (changes requested):
This pr is not an improve, it is a migration, but @Gokac9 you shouldn't add new behaviors in a migration.
Moreover, I'm not sure if this new behavior is appropiate for this module.

PR #622, opened 2024-12-27
[16.0][IMP] crm_claim: Improve search view

unaiberis:


anajuaristi (approved):

anajuaristi (approved):

anajuaristi (approved):

oihane (approved):
LGTM

crm_claim/views/crm_claim_views.xml
```diff
@@ -160,8 +160,8 @@
                         context="{'group_by':'stage_id'}"
                     />
                     <filter
-                        string="Type"
-                        name="type"
+                        string="Category"
+                        name="category"
```
pedrobaeza, line 164: Don't change this identifier, as it can be used in children modules.

anajuaristi: @pedrobaeza The module has an issue. It has been incorrect from begining and it's quite confusing for users. 
The string and name is TYPE but group_by is by categ_id. When they group by type they don't understand the behavior since the field "type" exists on claims and it's added by this module. 
It should be consistent. Category should group by category and type should group by type.
We are trying to change this months ago and it seems it's not going to be merged since our PRs are being closed everytime.
Please tell us how we must propose the PR to be merged. 

We only want to be consistent behavior on both claim fields
Thank you!

pedrobaeza: It has nothing to do with that, Ana. It's a question on stability. If you change the internal `name`, some modules will stop working. Just change the label, and for the user, the issue will be resolved.

PR #624, opened 2024-12-27
[16.0][IMP] crm_claim_type: Add search view

unaiberis:


anajuaristi (approved):

anajuaristi:
this is related to https://github.com/OCA/crm/pull/622
Label is wrong and impacting with this module. There is 2 categorization fields. Category coming from base and type coming from this module.
In 622, we are naming correctly the category field
In this one, we are including the field type on group by. 

@pedrobaeza would it be possible to merge them? They are really very basic PRs that IMHO are improving this branche. 
Please tell me if you could do it or I need to contact anyone.

Thank you very much

crm_claim_type/views/crm_claim.xml
```diff
@@ -14,4 +14,31 @@
...
+                    domain="[('claim_type', '!=', False)]"
+                />
+            </xpath>
+            <xpath expr="//group" position="inside">
+                <filter
+                    name="claim_type_group"
+                    string="Claim Type"
+                    context="{'group_by': 'claim_type'}"
+                />
+                <filter
```
pedrobaeza: This part shouldn't be in this module, as the stage is not added by `crm_claim_type`.

PR #545, opened 2024-01-31
[17.0][MIG] crm_project_task: Migration to 17.0

mide-dynapps:

PR #625, opened 2024-12-30
[MIG][18.0] crm_project_task (from 16.0)

acsonefho:
Migration of `crm_project_task` from 16.0


gurneyalex (approved):

acsonefho:
@OCA/crm-sales-marketing-maintainers Is it possible to reopen please :pray: 
Already in production.

SamyBrahimi (approved):

SamyBrahimi:
Really usefull to track the cost of a presale effort

crm_project_task/models/crm_lead.py
```diff
@@ -12,13 +13,20 @@ class CrmLead(models.Model):
...
     @api.depends("task_ids")
     def _compute_task_count(self):
+        domain = [("lead_id", "!=", False)]
+        data = self.env["project.task"].read_group(
+            domain, fields=["lead_id"], groupby=["lead_id"]
+        )
+        result = {d.get("lead_id")[0]: d.get("lead_id_count") for d in data}
         for lead in self:
-            lead.task_count = len(lead.task_ids)
+            lead.task_count = result.get(lead.id, 0)
```
yajo, line 22: issue: This change is introducing a few performance problems.

The first one is that you're searching for `[("lead_id", "!=", False)]`. This means that you will always obtain all the tasks in the database that are related to any lead. However, we only care about the ones linked to the current recordset.

The second problem is that you're forcing Odoo to flush data and issue at least one query to Postgres always. Yes, that might seem not much, but the previous implementation could live entirely on cached data and avoid any other queries.

So please would you undo this change?

crm_project_task/views/crm_lead.xml
```diff
@@ -14,7 +14,7 @@
                     type="action"
                     class="oe_highlight"
                     context="{'default_task_name': name}"
-                    attrs="{'invisible': ['|', ('active','=',False), ('task_count', '!=', 0)]}"
+                    invisible="not active or task_count != 0"
```
yajo, line 17: issue: Please migrate from v17 instead.

This change (and many others) was already done in https://github.com/OCA/crm/pull/545. Besides, that PR included some enhancements such as new demo data.

Could you please restart the migration, starting from the 17.0 branch instead? Thanks!

yajo (changes requested):
Some bugs introduced, and the migration should be done from v17.

@moduon MT-10809

crm_project_task/views/crm_lead.xml
acsonefho: Hello! Thanks for your review :muscle: 
The purpose of this PR is just to migrate the module.
Others improvements should be done into a dedicated PR. Because the migration could be approved and not the improvement

crm_project_task/wizards/crm_create_task.py
```diff
@@ -0,0 +1,82 @@
...
+                    "Please, create task from lead/opportunity."
+                )
+            )
+
+        # Create task
+        task = (
+            self.env["project.task"].sudo().create(self._get_data_create(lead, project))
+        )
+        # Messages in chatter
+        task.message_post(
```
EmilioPascual, line 44: Messages in chatter was fixed in v17 https://github.com/OCA/crm/pull/545. Can you review them?

EmilioPascual (changes requested):
Thank you for your job @acsonefho.
IMHO: The migration would be restarted from version 17. WDYT?

crm_project_task/views/crm_lead.xml
EmilioPascual: You're right but the migration should work ok. Some changes were made in version 17, if you don't add them then they will be lost.

rafaelbn:
Hi @acsonefho @gurneyalex @EmilioPascual @yajo

Thanks for your contributions and efforts here. I’d like to share a product perspective to help us decide the best way forward.

This PR was opened in December 2024 to migrate `crm_project_task` from 16.0 to 18.0. However, since then, the 17.0 version has received several improvements and bug fixes that are not included in this 18.0 PR.

From a product continuity point of view, any migration to a newer Odoo version should always reflect the most updated and validated version of the module, not just the closest available branch. Otherwise, we lose consistency, traceability and may even introduce regressions.

Unfortunately, OCA doesn’t yet have an automatic forward-porting bot like Odoo does, so we need to be careful with this manually.

My suggestion would be If @acsonefho doesn’t have capacity to update this PR, perhaps one of us could help finalize or recreate it — we’re happy to collaborate. Maybe close or supersede this PR with a new one that migrates from 17.0 to 18.0, ensuring all functional improvements are preserved.

Would that be okay for everyone? The goal is just to keep the migration process efficient, coherent, and product-friendly.

Thanks again for your work and dedication.

🤗 ❤️

acsonefho:
> Hi @acsonefho @gurneyalex @EmilioPascual @yajo
> 
> Thanks for your contributions and efforts here. I’d like to share a product perspective to help us decide the best way forward.
> 
> This PR was opened in December 2024 to migrate `crm_project_task` from 16.0 to 18.0. However, since then, the 17.0 version has received several improvements and bug fixes that are not included in this 18.0 PR.
> 
> From a product continuity point of view, any migration to a newer Odoo version should always reflect the most updated and validated version of the module, not just the closest available branch. Otherwise, we lose consistency, traceability and may even introduce regressions.
> 
> Unfortunately, OCA doesn’t yet have an automatic forward-porting bot like Odoo does, so we need to be careful with this manually.
> 
> My suggestion would be If @acsonefho doesn’t have capacity to update this PR, perhaps one of us could help finalize or recreate it — we’re happy to collaborate. Maybe close or supersede this PR with a new one that migrates from 17.0 to 18.0, ensuring all functional improvements are preserved.
> 
> Would that be okay for everyone? The goal is just to keep the migration process efficient, coherent, and product-friendly.
> 
> Thanks again for your work and dedication.
> 
> 🤗 ❤️

Hello! I agree with you @rafaelbn . My PR should be closed and a new one (based on 17) should be created.
Unfortunately, I don't have time now to do it.

yajo:
> Unfortunately, I don't have time now to do it.

Don't worry, we understand. We will take over that job then. Thanks!

PR #631, opened 2025-01-31
[16.0][IMP] crm_project_create: Archive or reactive the project and its analytical account from lead.

edlopen:
Hello,
This pull request adds the functionality that when archiving or unarchiving an opportunity, its associated projects and analytical accounts are also archived or unarchived. It also works when a lead is lost.

TODO: video

@moduon MT-8265

### DIFF
--- a/crm_project_create/models/crm_lead.py
+++ b/crm_project_create/models/crm_lead.py
@@ -8,3 +8,11 @@ class CrmLead(models.Model):
     _inherit = "crm.lead"
 
     project_id = fields.Many2one("project.project", string="Project")
+
+    def toggle_active(self):
+        """Archive or reactivate the project and their analytic account on lead toggle."""
+        res = super().toggle_active()
+        for lead in self.filtered(lambda l: l.project_id):
+            lead.sudo().project_id.active = lead.active
+            lead.sudo().project_id.analytic_account_id.active = lead.active
+        return res
--- a/crm_project_create/tests/test_crm_create_project.py
+++ b/crm_project_create/tests/test_crm_create_project.py
@@ -38,9 +38,12 @@ class TestCrmCreateProject(TransactionCase):
                 "user_id": cls.user_salesman.id,
             }
         )
+        cls.default_plan = cls.env["account.analytic.plan"].create(
+            {"name": "Default", "company_id": False}
+        )
 
     @users("user_salesman")
-    def test_crm_create_project(self):
+    def test_crm_project_create(self):
         """Test the creation of a project from a lead."""
         wizard_form = Form(
             self.env["crm.create.project"].with_context(
@@ -51,7 +54,6 @@ class TestCrmCreateProject(TransactionCase):
             )
         )
         self.assertEqual(wizard_form.project_name, self.lead.name)
-
         wizard_form.project_name = "Test Project"
         wizard_form.project_description = "Test Description"
         wizard = wizard_form.save()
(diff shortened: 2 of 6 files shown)


crm_project_create/models/crm_lead.py
```diff
@@ -8,3 +8,11 @@ class CrmLead(models.Model):
     _inherit = "crm.lead"
 
     project_id = fields.Many2one("project.project", string="Project")
+
+    def toggle_active(self):
+        """Archive or reactivate the project and their analytic account on lead toggle."""
+        for lead in self.filtered(lambda l: l.project_id):
+            lead_state = lead.active
+            lead.sudo().project_id.active = not lead_state
+            lead.sudo().project_id.analytic_account_id.active = not lead_state
```
EmilioPascual:
```suggestion
            lead.sudo().project_id.active = not lead.active
            lead.sudo().project_id.analytic_account_id.active = not lead.active
```

crm_project_create/tests/test_crm_lead.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2025 Moduon Team S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
+
+from odoo.tests.common import Form, users
+
+from .common import TestCrmLeadCommon
+
+
+class TestCrmLead(TestCrmLeadCommon):
```
EmilioPascual: I think this file is not necessary, you are repeating the method `test_crm_lead_lost` from the file `crm_project_create/tests/test_crm_create_project.py` and method `test_crm_lead_archive` should be implemented in the other file.

I think you can implement the checks in the same method test_crm_create_project, in this way the new files `crm_project_create/tests/test_crm_lead.py` and `crm_project_create/tests/common.py` aren't neccesary.

EmilioPascual (changes requested):

EmilioPascual:
Also you should review the tests.

crm_project_create/tests/test_crm_create_project.py
```diff
@@ -63,3 +65,30 @@ def test_crm_create_project(self):
...
+        analytic_account = self.env["account.analytic.account"].create(
+            {
+                "name": "Test analytic account",
+                "plan_id": self.default_plan.id,
+                "company_id": False,
+            }
+        )
+        self.lead.project_id.analytic_account_id = analytic_account
+        self.assertTrue(self.lead.project_id.active)
+        self.assertTrue(self.lead.project_id.analytic_account_id.active)
```
EmilioPascual:
```suggestion
```
Why do you check the project and the analytic account are active? I think it's not necessary, the project and the account analytic have just been created, they should be active unless you have added a new behavior for it, but this isn't the case.

EmilioPascual (changes requested):
A little change

EmilioPascual (approved):
Code and functional review.

Good Job @edlopen!!

crm_project_create/models/crm_lead.py
```diff
@@ -8,3 +8,10 @@ class CrmLead(models.Model):
     _inherit = "crm.lead"
 
     project_id = fields.Many2one("project.project", string="Project")
+
+    def toggle_active(self):
+        """Archive or reactivate the project and their analytic account on lead toggle."""
+        for lead in self.filtered(lambda l: l.project_id):
+            lead.sudo().project_id.active = not lead.active
+            lead.sudo().project_id.analytic_account_id.active = not lead.active
+        return super().toggle_active()
```
Shide: Why not move `super().toggle_active()` to top and avoid the confusing `not lead.active`?

```suggestion
        res = super().toggle_active()
        for lead in self.filtered(lambda l: l.project_id):
            lead.sudo().project_id.active = lead.active
            lead.sudo().project_id.analytic_account_id.active = lead.active
        return res
```

Shide (commented):
**nitpick**: Avoid negate variables when not needed

crm_project_create/models/crm_lead.py
edlopen: It makes sense. Thanks for the advice!

rafaelbn (approved):
👍🏼 tested functionally in runbot

PR #640, opened 2025-04-08
[15.0][IMP] crm_multicompany_reporting_currency

imlopes:
* Use the current company currency when summing amounts on CRM kanban.

If we use more than one currency on `crm.lead`, odoo will sum the `expected_revenue` ignoring the currency : 
USD 50,00 + EUR 50,00 = USD 100,00 (odoo takes the first record on kanban view)

With this update, we will use the `current_company_currency` : 
Company currency = BRL
(1 USD = 1.5 BRL and 1 EUR = 2 BRL)

USD 50,00 = 75,00 BRL
EUR 50,00 = 100,00 BRL

We display on kanban progress bar `BRL 175,00`

The second commit is the same than here: https://github.com/OCA/crm/pull/642

### DIFF
--- a/crm_multicompany_reporting_currency/models/crm_lead.py
+++ b/crm_multicompany_reporting_currency/models/crm_lead.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
 from odoo import api, fields, models
+from odoo.fields import Date
 
 
 class CrmLead(models.Model):
@@ -36,6 +37,15 @@ class CrmLead(models.Model):
         index=True,
         readonly=True,
     )
+
+    # technical field to be used in sum_field options on kanban
+    current_company_currency = fields.Many2one(
+        "res.currency", compute="_compute_current_company_currency", readonly=True
+    )
+
+    def _compute_current_company_currency(self):
+        for lead in self:
+            lead.current_company_currency = self.env.company.currency_id
 
     @api.depends("company_currency")
     def _compute_multicompany_reporting_currency_id(self):
@@ -81,3 +91,33 @@ class CrmLead(models.Model):
             else:
                 to_amount = record.expected_revenue * record.currency_rate
             record.amount_multicompany_reporting_currency = to_amount
+
+    @api.model
+    def read_group(self, domain, fields, groupby, **kwargs):
+        # Override read_group to calculate the sum of the
+        # expected revenue in current company currency
+        current_company = self.env.company
+        result = super().read_group(
+            domain,
+            fields,
+            groupby,
+            **kwargs,
+        )
+        opportunities = self.env["crm.lead"]
+        for line in result:
+            if "__domain" in line:
+                opportunities = self.search(line["__domain"])
+
+            if "expected_revenue" in fields:
+                global_revenue = sum(
+                    opportunities.mapped(
+                        lambda opp: opp.company_currency._convert(
+                            opp.expected_revenue,
+                            current_company.currency_id,
+                            current_company,
+                            Date.today(),
+                        )
+                    )
+                )
+                line["expected_revenue"] = global_revenue
+        return result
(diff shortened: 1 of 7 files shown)


crm_multicompany_reporting_currency/static/src/js/views/kanban/kanban_column_progressbar.js
```diff
@@ -0,0 +1,27 @@
...
+                this._super.apply(this, arguments);
+                var companyCurrency =
+                    columnState.progressBarValues.company_currency_field;
+                if (companyCurrency && columnState.data.length) {
+                    this.currency =
+                        session.currencies[
+                            columnState.data[0].data[companyCurrency].res_id
+                        ];
+                }
+            },
```
yankinmax, line 24: I suppose `companyCurrency` field existence is to incapsulate the usage only on CRM? Maybe there is more elegant way to check if current active model has such an attribute?

crm_multicompany_reporting_currency/models/crm_lead.py
```diff
@@ -81,3 +91,33 @@ def _compute_amount_multicompany_reporting_currency(self):
             else:
                 to_amount = record.expected_revenue * record.currency_rate
             record.amount_multicompany_reporting_currency = to_amount
+
+    @api.model
+    def read_group(self, domain, fields, groupby, **kwargs):
+        # Override read_group to calculate the sum of the
+        # expected revenue in current company currency
+        current_company = self.env.company
+        result = super().read_group(
```
yankinmax, line 100: If we use `read_group` to calculate the sum, why do we use `search()` and `sum()` then? Can't we try with `read_group` aggregation?

crm_multicompany_reporting_currency/models/crm_lead.py
```diff
@@ -37,6 +38,15 @@ def _get_multicompany_reporting_currency_id(self):
         readonly=True,
     )
 
+    # technical field to be used in sum_field options on kanban
+    current_company_currency = fields.Many2one(
```
SilvioC2C, line 42: `Many2one` fields should conventionally end up with `_id`
```suggestion
    current_company_currency_id = fields.Many2one(
```

crm_multicompany_reporting_currency/models/crm_lead.py
```diff
@@ -37,6 +38,15 @@ def _get_multicompany_reporting_currency_id(self):
...
+        "res.currency", compute="_compute_current_company_currency", readonly=True
+    )
+
+    def _compute_current_company_currency(self):
```
SilvioC2C, line 46: I would add `@api.depends_context("company")`

crm_multicompany_reporting_currency/models/crm_lead.py
```diff
@@ -81,3 +91,33 @@ def _compute_amount_multicompany_reporting_currency(self):
...
+        result = super().read_group(
+            domain,
+            fields,
+            groupby,
+            **kwargs,
+        )
+        opportunities = self.env["crm.lead"]
+        for line in result:
+            if "__domain" in line:
+                opportunities = self.search(line["__domain"])
```
SilvioC2C, line 109: If a line doesn't define the `__domain` keyword, then it will use the opportunities of the previous line, which is wrong. Please fix this.

SilvioC2C (changes requested):
I see an inconsistency in the currency usage, as the field defined on `crm.lead` records is used in the JS code, but the `read_group` uses the environmental currency. Can you please uniform the behavior?

Also, I see there's already another PR opened by you for fixing `crm_salesperson_planner`, you should rebase this PR upon that instead of adding another commit to fix the same issue twice.

imlopes:
Closing this PR as finally I think this module do almost the same things (with few differences) https://github.com/OCA/crm/tree/14.0/crm_lead_currency, which is available on 14.0, 16.0 and 17.0

PR #644, opened 2025-04-22
[18.0][MIG] crm_lead_product: Migration to 18.0

joelkm:
#600


joelkm:
I see that there was a previous pr #608 but it wasn't recognized by the OCA bot. Idk if it is abandoned or not, what do we do in this case?

etobella:
@joelkm In this case I would recommend a review on the original PR in order to speed up (I will make a review on #608 in order to help). If we have 2 approvals, we can merge it.

PR #647, opened 2025-04-29
[16.0][ADD] crm_partner_required: New module

EmilioPascual:
It makes the field partner required in the opportunity

¿@rafaelbn @yajo @Gelojr @fcvalgar can you review it? Thanks.

MT-10018 @moduon


rafaelbn:
Thank you very much @EmilioPascual !

I miss the loom video 😉 

Here you are, could you review please?

https://www.loom.com/share/9bf18dfff6564593828d5895a4403a05?sid=7139ad6d-27f1-4f1e-8499-cef9bcd34178

I agree not adding more fields in the kanban view for better UX, and maybe I also agree **if intended** not mandatory the field in the kanban view. This is a question 😄 

With ❤️ , RB.

EmilioPascual:
> I agree not adding more fields in the kanban view for better UX, and maybe I also agree **if intended** not mandatory the field in the kanban view. This is a question 😄

I was thinking about it and perhaps the partner should also be mandatory in quick create form. I've pushed new changes. @rafaelbn can you review again?

crm_partner_required/tests/test_crm_partner_required.py
```diff
@@ -0,0 +1,38 @@
...
+        opportunity_form = Form(
+            self.env["crm.lead"].with_context(default_type="opportunity")
+        )
+        opportunity_form.name = "Test Opportunity"
+        with self.assertRaises(AssertionError):
+            opportunity_form.save()
+        opportunity_form.partner_id = self.partner
+        opportunity_form.save()
+        # Check required in opportunity in quick create form
+        opportunity_quick_create_form = Form(
```
yajo, line 26: issue: IMHO this is a separate unit test. One for each form.

crm_partner_required/tests/test_crm_partner_required.py
```diff
@@ -0,0 +1,38 @@
...
+        )
+        opportunity_quick_create_form.name = "Test Opportunity Quick Create"
+        with self.assertRaises(AssertionError):
+            opportunity_quick_create_form.save()
+        opportunity_quick_create_form.partner_id = self.partner
+        opportunity_quick_create_form.save()
+        # Check required in lead in default form
+        lead_form = Form(self.env["crm.lead"].with_context(default_type="lead"))
+        lead_form.name = "Test Lead"
+        lead_form.save()
```
yajo, line 38: issue: and this should be a separate test also.

yajo (changes requested):
> the partner should also be mandatory in quick create form

AFAICS this is already fixed, thanks!

crm_partner_required/tests/test_crm_partner_required.py
EmilioPascual: Updated. Thanks for comments 🤗

crm_partner_required/tests/test_crm_partner_required.py
EmilioPascual: Done

yajo (approved):

rafaelbn (approved):
👍🏼 OK!

PR #648, opened 2025-05-04
[18.0][MIG] crm_lead_product: Migration 18.0

joelkm:


joelkm:
@etobella , I was hoping that the original contributor would fix the pr but it urges me to have this module in v18. I'm not used to contributing so instead of superseeding I've done the migration again. Is this good? Also, why is the codecov check failing? Did I miss some step? Thanks!!

etobella:
Well, it would be better to keep his commit. If it too harsh, you can edit your commit and add him as coauthor. However, if in a few days we don't receive a comment we can proceed with this one

etobella (approved):

rompelhd (approved):

giarve (approved):

joelkm:
Thanks for your help @etobella !!!

PR #650, opened 2025-05-13
[IMP] crm_lead_product: We should filter for sale products only

etobella:
CRM is part of the sale process. It makes no sense to add products that cannot be sold


Abranes (approved):

jaenbox (approved):
LGTM

PR #658, opened 2025-06-28
[18.0][MIG] crm_won_reason: Migration to 18.0

apami0111:
- Migrated from 15.0 to 18.0.
- Replaced deprecated `attrs` with new `invisible` expression in `crm_won_reason/views/crm_views.xml`:
- Updated test class inheritance from "TransactionCase" to "BaseCommon".


BhaveshHeliconia (changes requested):
Two "Won" buttons are being displayed.
Screenshot from 2025-07-30 15-55-22

apami0111:
@BhaveshHeliconia Thank you for the feedback. I have reviewed the case and confirm that the crm_won_reason module by itself does not present the issue of two "Won" buttons. This behavior only occurs when it is installed together with the crm_won_restrict_per_stage module, as both apply an invisible attribute to the same button, causing the view to be triggered and resulting in two instances of the button being displayed.
In an environment with only crm_won_reason, both functionality and interface work as expected.
Captura desde 2025-08-07 21-44-26

BhaveshHeliconia:
@apami0111 : The functionality works perfectly, but having the same "Won" label for both buttons can be confusing.

celm1990 (approved):
LGTM 

@pedrobaeza could you review this?

crm_won_reason/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+
+{
+    "name": "CRM won reason",
+    "version": "18.0.1.0.0",
+    "category": "Customer Relationship Management",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/crm",
+    "license": "AGPL-3",
+    "depends": [
+        "base",
```
jaenbox: depends 'base' is not necessary

apami0111: Thank you very much for your observation! @jaenbox 
I've already made the change.

jaenbox (approved):
LGTM

celm1990:
ping @pedrobaeza

pedrobaeza:
@BhaveshHeliconia has requested changes

apami0111:
@pedrobaeza Thanks for the note. The issue mentioned in
> @BhaveshHeliconia has requested changes
is already resolved. As I explained here:
https://github.com/OCA/crm/pull/658#issuecomment-3177391757

the duplicated “Won” button was not caused by crm_won_reason, but by its interaction with crm_won_restrict_per_stage, since both apply an invisible attribute to the same button.

With only crm_won_reason installed, everything works correctly.

pedrobaeza:
Yeah, but he should confirm it.

BhaveshHeliconia (approved):

apami0111:
ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge nobump

PR #664, opened 2025-07-30
[18.0][ADD] marketing_crm_calendar: copy campaign fields from phonecalls to calendar events

BetoRaja:
@HaraldPanten @javierizaca 
This module extends calendar events by adding UTM tracking fields: campaign, source, and medium. It links calendar events to related CRM phonecalls automatically when created from a phonecall context, inheriting their UTM values. The calendar event views are customized to display these tracking fields in the form and tree views for better marketing analysis.

T-8642


HaraldPanten (changes requested):
Module name needs to be improved

javierizaca (changes requested):
Hi @BetoRaja  

The foundation of the approach works correctly, although I think it would be desirable—so as to improve the user experience—to modify the following:

1. The order of the fields in the events’ form view: Campaign > Source > Medium. This follows the standard that Odoo uses for other models, since it is the way to hierarchize data in Marketing.

image

2. When we create the Calendar entry from a Call, the Call’s Campaign, Medium, and Source data are not copied directly into the draft; instead, they only show up after saving the record.

This means the user sees empty fields and might think they need to fill them in manually—if their permissions allow it. Would it be possible to transfer these values upon creation of the record, without having to save it first?

image

Thank you very much!

javierizaca (approved):
At a functional level everything works perfectly, the name of the module seems adequate to me since it aligns with the existing marketing_crm_partner

ValentinVinagre (changes requested):
These changes have been pushed to the main module, which isn't correct. Everything should be in the extension module.

PR #666, opened 2025-08-05
[MIG][18.0] crm_project_task

u0f:
Straight forward migration

MT-10809 @moduon @sabrinaRMartin @Xino61122 @yajo @EmilioPascual 

Configure the default project in CRM > Configuration > Settings > **Crm Default Project**
image

Crm > My Pipe line > Select and existing lead > **Create task** 
image

The assigned project is the one configured before
image

PR #667, opened 2025-08-05
[18.0][MIG] crm_project_task: Migration to 18.0

MohamedOsman7:
Replaces #625


yajo (changes requested):
As I explained in https://github.com/OCA/crm/pull/666#issuecomment-3154567671, **this module should not be merged**. `crm_lead_to_task` is essentially the same thing.

See both in action:

Image

Differences:


Feature | crm_project_task | crm_lead_to_task | Proposal
-- | -- | -- | --
Archive lead automatically | No | Yes | Add “Archive lead” checkbox to the wizard in `crm_lead_to_task`.
Configure a default project in settings | Yes | No | Add that feature to crm_lead_to_task.
Allow editing task and description directly through the wizard. | Yes | No | Probably this is unnecessary. You can edit that after creating the task.
Allow linking multiple tasks from a single lead. | Yes | Yes |  
Allow back-and-forth navigation between task and lead | Yes | Yes |  

As you can see, for the few feature gaps between both modules, a simple enhancement in the existing `crm_lead_to_task` should suffice. That would help more and be more maintainable than having 2 different implementations of the same thing.

Would you please close this PR and open those enhancement if needed?

Thanks!

glwr:
@yajo 
Thank you again for clarifying.
And we agree with you, the actual part of creating a task and linking the two models is the same.

We are now adding the following functions to the “crm_lead_to_task” module:

1. “Archive lead” = yes/no configuration. This allows you to specify in the settings whether a lead should be archived. Depending on this, the button would then be called “Convert to task” or “Create task”. (I would name the buttons this way for user-friendliness and recognition)

2. “Default project” configuration. This allows you to define a default project to be used in the settings. If none is stored, the dialog would ask for a project.

Thus we have 2 buttons that appear with different names depending on the setting and in one case archive the lead and in the other case leave the lead untouched, but call up the same dialog or execute the same action of task creation and linking.
The dialog to select a project would then be displayed depending on the second setting, the “Default project”.
We leave out the task name and description, we don't need that either.

yajo:
OK! Looks good then!



> 2\. “Default project” configuration. This allows you to define a default project to be used in the settings. If none is stored, the dialog would ask for a project.

If that's the meaning, then I'd label it "force project", because you're not adding a default value, but instead using that always no matter what.

I'll close this PR to let you focus in those enhancements.

Thanks for your understanding 😊

PR #672, opened 2025-08-16
[17.0][ADD] crm_stage_project_task

geomer198:
This module bridges CRM pipeline management with project tasks, automatically generating predefined tasks when leads/opportunities transition between stages. It enforces process compliance by optionally requiring task completion before stage progression.

Core Functionality:

- Configurable task templates per CRM stage

- Optional stage transition validation (blocks progression if tasks incomplete)

- Integrated task management within CRM form view

Key Features:

- Default project configuration for tasks

- Stage-specific task templates

- Task state validation system


Bearnard21 (approved):
Functionlity LGTM

Bearnard21:
@rafaelbn Please take a look, thank you

Bearnard21:
@rvalyi @azmeuk @pedrobaeza  could you please take a look
Thank you

azmeuk:
Why me?

geomer198:
@ivs-cetmix Make a code review please.

crm_stage_project_task/readme/CONTEXT.md
```diff
@@ -0,0 +1,35 @@
...
+
+2. No Process Enforcement
+    
+    - Users can skip mandatory steps (e.g., contract reviews) by moving stages prematurely
+    - Hard to track which stage-specific actions are pending
+
+3. Disconnected Workflows
+    
+    - Tasks exist in Projects but lack visibility in CRM
+    - No native link between pipeline stages and task templates
```
ivs-cetmix, line 20: Hello @geomer198 thank you for you contribution!

Could you please clarify why do you need to link the CRM lead directly to a dedicated project if you need, why cannot this be done using the standard `CRM -> Sales -> Project` flow?
If you have a complex lead closure process that has multiple tasks to be completed, then  you can create a quotation from the opportunity, link a project to that quotation and later use that link to bind tasks with the lead.
Regarding `Teams waste time recreating repetitive tasks (e.g., "Send proposal") for every lead` cannot this be done with activities?
Also have a look at this module, probably you could use it as a basis and extend it instead: https://github.com/OCA/crm/tree/17.0/crm_project_task

ivs-cetmix (changes requested):

crm_stage_project_task/readme/CONTEXT.md
geomer198: @ivs-cetmix, thank you for your comment! This solution has been used in one of our projects. Therefore, we decided to publish the module so that other users could also take advantage of its capabilities.

Bearnard21: @ivs-cetmix Every task in the CRM flow, such as sending a proposal, might require multiple additional actions and discussions among different teams, with confidential attachments used only by the particular users.  Also, we might assign a task to an external department user who must not see the CRM flow and other commercial info it may contain. Therefore, the solution is to create a configurable separate task for each stage of the CRM flow.

Bearnard21:
@rafaelbn Could you please take a look?

PR #674, opened 2025-08-22
[18.0][MIG] crm_industry: Migration to v18.0

sabrinaRMartin:
Migrated to version 18

There have been no significant changes.

@moduon MT-11100

@mmrondon 

https://www.loom.com/share/500507824ac04447a264ef8554b65265?sid=2bbbb7e7-c2c7-439b-bf6f-68f3e752c094

PR #675, opened 2025-09-01
[18.0][IMP] crm_industry: Added new test

sabrinaRMartin:
Added new tests for better control of crm_industry functionality, this also fixes Codecov errors.

https://github.com/OCA/crm/pull/674

@moduon @mmrondon 

MT-11100

### DIFF
--- a/crm_industry/tests/test_crm_lead.py
+++ b/crm_industry/tests/test_crm_lead.py
@@ -66,3 +66,65 @@ class TestCrmLead(TransactionCase):
         lead = lead_form.save()
         self.assertEqual(lead.industry_id, customer.industry_id)
         self.assertEqual(lead.secondary_industry_ids, customer.secondary_industry_ids)
+
+        def test_override_propagate_industries_from_contact(self):
+            res_partner_industry = self.env["res.partner.industry"].create(
+                {"name": "Parent Industry"}
+            )
+            industry_a, industry_b = res_partner_industry.create(
+                [
+                    {"name": "IT/Communications", "parent_id": res_partner_industry.id},
+                    {
+                        "name": "AI/Machine Learning",
+                        "parent_id": res_partner_industry.id,
+                    },
+                ]
+            )
+
+            customer = self.env["res.partner"].create(
+                {
+                    "name": "Test Customer",
+                    "industry_id": res_partner_industry.id,
+                    "secondary_industry_ids": [(6, 0, [industry_a.id, industry_b.id])],
+                }
+            )
+
+            lead_form = Form(self.env["crm.lead"])
+            lead_form.name = "Lead from partner"
+            lead_form.partner_id = customer
+            lead = lead_form.save()
+
+            self.assertEqual(lead.industry_id, customer.industry_id)
+            self.assertEqual(
+                lead.secondary_industry_ids, customer.secondary_industry_ids
+            )
+
+        def test_create_with_partner_does_not_override_explicit_vals(self):
+            industry_test = self.env["res.partner.industry"].create({"name": "Test"})
+            industry_1, industry_2, industry_3 = self.env[
+                "res.partner.industry"
+            ].create(
+                [
+                    {"name": "Test 01", "parent_id": industry_test.id},
+                    {"name": "Test 02", "parent_id": industry_test.id},
+                    {"name": "Test 03", "parent_id": industry_test.id},
+                ]
+            )
+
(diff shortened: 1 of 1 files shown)


crm_industry/tests/test_crm_lead.py
```diff
@@ -66,3 +66,61 @@ def test_propagate_industries_from_contact(self):
...
         self.assertEqual(lead.secondary_industry_ids, customer.secondary_industry_ids)
+
+        def test_override_propagate_industries_from_contact(self):
+            res_partner_industry = self.env["res.partner.industry"]
+            industry_a = res_partner_industry.create(
+                {"name": "IT/Communications", "parent_id": res_partner_industry.id}
+            )
+            industry_b = res_partner_industry.create(
+                {"name": "AI/Machine Learning", "parent_id": res_partner_industry.id}
+            )
```
yajo: **question(blocking):** what were you trying to do here with the parent industry?

I mean... when you do `res_partner_industry = self.env["res.partner.industry"]`, you will get an empty recordset. Then when you get its `.id`, you'll get `False`. So, this code is equivalent to:


```suggestion
            industry_a, industry_b = self.env["res.partner.industry"].create(
                [
                    {"name": "IT/Communications"},
                    {"name": "AI/Machine Learning"},
                ]
            )
```

But it seems like this is not what you wanted, because you probably want an industry in place for the assertions you have below... So, maybe you meant this?:


```suggestion
            res_partner_industry = self.env["res.partner.industry"].create({"name": "Parent"})
            industry_a, industry_b = self.env["res.partner.industry"].create(
                [
                    {"name": "IT/Communications", "parent_id": res_partner_industry.id},
                    {"name": "AI/Machine Learning", "parent_id": res_partner_industry.id},
                ]
            )
```

crm_industry/tests/test_crm_lead.py
```diff
@@ -66,3 +66,61 @@ def test_propagate_industries_from_contact(self):
...
+                    "secondary_industry_ids": [(6, 0, [industry_a.id, industry_b.id])],
+                }
+            )
+
+            lead = self.env["crm.lead"].create(
+                {
+                    "name": "Lead from partner",
+                    "partner_id": customer.id,
+                }
+            )
```
yajo: **suggestion:** Please use [the `Form` class](https://www.odoo.com/documentation/18.0/developer/reference/backend/testing.html#odoo.tests.Form) here, to imitate human interaction and make sure it works as expected. That will make the tests more meaningful.

crm_industry/tests/test_crm_lead.py
```diff
@@ -66,3 +66,61 @@ def test_propagate_industries_from_contact(self):
...
+
+            lead = self.env["crm.lead"].create(
+                {
+                    "name": "Test lead",
+                    "partner_id": customer.id,
+                    "industry_id": industry_1.id,
+                    "secondary_industry_ids": [(6, 0, [industry_2.id])],
+                }
+            )
```
yajo: **suggestion:** here use the `Form` class too.

yajo (changes requested):

crm_industry/tests/test_crm_lead.py
sabrinaRMartin: ups! 🫣Yes, what I wanted to do was to define a parent industry so that I could verify that this was fulfilled, the spread of the industry from the partner to the lead:
`if customer.industry_id and not vals.get("industry_id"):
      vals.update({"industry_id": customer.industry_id.id})`

crm_industry/tests/test_crm_lead.py
sabrinaRMartin: Applied use of Form

crm_industry/tests/test_crm_lead.py
sabrinaRMartin: Applied use of Form

EmilioPascual (approved):
Code review

yajo (changes requested):
Thanks, it looks excellent now!

However, I see that we have 2 commits here for the same thing. Would you please [squash them](https://stackoverflow.com/a/5189600/1468388) into one, to be able to merge? [Currently I can't ask ocabot to squash on merge](https://github.com/OCA/oca-github-bot/issues/43).

sabrinaRMartin:
@yajo fixed!

yajo (approved):
/ocabot merge patch

PR #689, opened 2025-10-06
[IMP] crm_lead_to_task: Usability improves

u0f:
After merging “crm_project_task,” some usability issues were reported. The following improvements have been added:

- Unified to a button in the lead with a wizard that always appears.
- If a project is selected in the options, it will open by default in the wizard in read-only mode.
- Within the wizard, you can:
  - Convert lead to task: this is how it worked before, where you can archive the lead and move the chatter to the task.
  - Create task and link: a task linked to the lead will be created, but the chatter will not be moved and the lead will not be archived.
  - README updated
  - New tests to cover up this improves.

https://www.loom.com/share/142ac524c4274a75b6aef5d1ef6feb5b

@moduon MT-11907


rafaelbn (changes requested):
Hello, thank you for this improvements 🙏🏼 

- [ ] Please be sure we solve https://github.com/OCA/crm/issues/481
- [ ] Add Moduon as author of the module and maintain the contributor  and maintainers of https://github.com/OCA/crm/tree/16.0/crm_project_task

Thank you! 😄 ❤️

crm_lead_to_task/wizard/crm_lead_convert2task_views.xml
```diff
@@ -6,13 +6,22 @@
         <field name="arch" type="xml">
             <form string="Convert to Task">
```
yajo, line 7: issue: this form title is now misleading

crm_lead_to_task/models/crm_lead.py
```diff
@@ -79,16 +64,36 @@ def _action_create_and_open_task(self, project):
...
-            action = self.env.ref(
-                "crm_lead_to_task.crm_lead_convert2task_action"
-            ).read()[0]
-            action["context"] = {"default_lead_id": self.id}
+            context["default_project_id"] = self.company_id.crm_force_project_id.id
 
+        action["context"] = context
         return action
+
+    def _move_chatter(self, task):
```
yajo, line 79: issue: remember that many modules can coexist, so it's better if method names are more explicit:


```suggestion
    def _move_chatter_to_task(self, task):
```

crm_lead_to_task/models/crm_lead.py
```diff
@@ -79,16 +64,36 @@ def _action_create_and_open_task(self, project):
...
+                ("res_model", "=", "crm.lead"),
+                ("res_id", "=", self.id),
+            ]
+        ).write({"res_model": "project.task", "res_id": task.id})
+
+    def _convert_lead_to_task(self, project):
+        task = self._create_task_from_lead(project)
+        self._move_chatter(task)
+        if self.company_id.crm_archive_lead_on_convert:
+            self.active = False
```
yajo, line 94: question: does this make sense now? If users have the ability to click on "convert to task", shouldn't that just archive always?

IMHO the `crm_archive_lead_on_convert` should be an option with 3 choices:
- Always
- Never
- Empty (let the user choose)

crm_lead_to_task/views/crm_lead_views.xml
```diff
@@ -6,25 +6,15 @@
...
-                <!--Create a Task button (no archive) -->
+                <!--Create a Task action-->
                 <button
-                    name="action_crm_to_task"
+                    name="action_create_task"
                     id="action_create_task_from_lead"
                     type="object"
                     string="Create Task"
                     class="oe_stat_button"
                     groups="project.group_project_user"
```
yajo, line 16: issue: salespersons cannot convert to task if they don't have project permissions.

Image


```suggestion
```

Salespeople should be able to perform the conversion (not the linking).

yajo (changes requested):

PR #697, opened 2025-10-30
[16.0][FIX] crm_stage_multi_team: makes stage in form view work as normal + show stages if no leads

eduezerouali-tecnativa:
cc @Tecnativa TT58851

Makes the crm works same way it should whit no multi_team. It show only the stages related to the teams in form view and also in kanban view it shows the stage even if it has no leads. 

ping @pedrobaeza @victoralmau  could you review?


crm_stage_multi_team/models/crm_lead.py
```diff
@@ -45,14 +45,14 @@ def _read_group_stage_ids(self, stages, domain, order):
         )
         if test_condition or self.env.context.get("no_crm_stage_multi_team"):
             return super()._read_group_stage_ids(stages, domain, order)
-        team_id = self._context.get("default_team_id")
-        if team_id:
+        team_ids = self.env["crm.team"]._get_default_team_id().ids
```
pedrobaeza: Isn't this returning only one team?

eduezerouali-tecnativa:
```suggestion
        team_ids = self.env.user.crm_team_ids.ids
```
You are right. Could I get the teams direct from the user or should i override the method `_get_default_team_id`?

pedrobaeza: Don't change `_get_default_team_id`. This is something exclusive to this specific method.

pedrobaeza (approved):
Let's merge the bugfix:

/ocabot merge patch

pedrobaeza:
Please fw-port it to upper versions.

PR #661, opened 2025-07-23
[16.0] [ADD] crm_team_zip_assign: partner auto assign crm team by zip

rrebollo:
### DESCRIPTION

Auto-assign CRM teams to partners based on ZIP code patterns using regular expressions.

Check out the README.rst for a complete overview.

@BinhexTeam

PR #709, opened 2025-11-12
[17.0] [FIX] crm_phonecall: Permission error in action

Tomero-Dev:
Viewing calls from the CRM Lead results in a permissions error caused by the action being read.
The issue has been resolved by updating to the _for_xml_id() format, which is the current standard in this version.

- Impacted versions:

17.0

- Steps to reproduce:

Go to the CRM -> one record -> open form view
Click the Call SmartButton with a user who is not an administrator and does not have Settings permissions.

### DIFF
--- a/crm_phonecall/models/crm_lead.py
+++ b/crm_phonecall/models/crm_lead.py
@@ -25,7 +25,9 @@ class CrmLead(models.Model):
 
     def button_open_phonecall(self):
         self.ensure_one()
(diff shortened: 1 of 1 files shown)


TadeuszKarpinski (approved):

rrebollo (changes requested):
Code Review. LGTM! but the commit messages should include targeted addon. Something along [17.0] [FIX] crm_phonecall: ...

rrebollo:
Would you be so kind to review mine back? #661

PR #710, opened 2025-11-14
[FIX] crm_security_group: load_data goes with env, not cr

etobella:
@victoralmau @pedrobaeza


pedrobaeza (approved):
It was `cr` in previous version, but not on 18...

PR #711, opened 2025-11-19
[FIX] Lead to Task: No archive fix

jasiel-osi:
Making a fix to lead_to_task to avoid been archive, and coping the chatter information to the task


rrebollo:
PR name should be aligned with repository conventions. Commit history messages are not aligned with the OCA conventions either.

PR #712, opened 2025-11-28
[19.0][MIG] crm_phonecall: migration to 19.0

eugenios73:
migration crm_phonecall from 18.0 to 19.0

The pre-commit-config update in commit [[19.0][FIX] update pre-commit-config](https://github.com/OCA/crm/pull/712/commits/bd7ecc0b44f451549116b255ba9de3589500a53f) is to resolve the "user_id" error issue that was shown in the last pre-commit command and which I commented on in this message https://github.com/OCA/crm/pull/712#issuecomment-3595497131

PR #724, opened 2026-01-21
[19.0][FIX] update pre-commit-config

eugenios73:
The update to pre-commit-config resolves the issue with the command, which appears to be outdated due to the change in version 19.0 to the "user_ids" attribute of the "ir.filters" base model. This attribute has changed compared to version 18.0, where it was "user_id". The commit showing this change can be found here: https://github.com/odoo/odoo/commit/414e55cf7c3971a7ba0a7c96855db4c23e77a5c4

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,10 +1,9 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.35
+_commit: v1.42
 _src_path: git+https://github.com/OCA/oca-addons-repo-template
 additional_ruff_rules: []
-ci: GitHub
 convert_readme_fragments_to_markdown: true
-enable_checklog_odoo: true
+enable_checklog_odoo: false
 generate_requirements_txt: true
 github_check_license: true
 github_ci_extra_env: {}
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,3 +1,4 @@
+
 name: pre-commit
 
 on:
@@ -16,7 +17,9 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.12"
+          cache: 'pip'
+          cache-dependency-path: '.pre-commit-config.yaml'
       - name: Get python version
         run: echo "PY=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
       - uses: actions/cache@v4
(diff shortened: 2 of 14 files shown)


Reyes4711-S73 (approved):
LGTM

sbiosca-s73 (approved):

Reyes4711-S73:
@OCA/crm-sales-marketing-maintainers Please, can you review/merge this PR?

luisDIXMIT (approved):
LGTM!

eugenios73:
@pedrobaeza Please can you review / mergtte this PR ?

pedrobaeza (changes requested):
Please don't do this manually, but run copier update.

eugenios73:
Hi again @pedrobaeza , I've already applied the changes you mentioned and fixed some new errors in the repository modules that weren't showing up before the update. As for the error appearing in the tests, I've already solved it in PR https://github.com/OCA/crm/pull/712. Thanks in advance.

checklog-odoo.cfg
```diff
@@ -1,3 +0,0 @@
-[checklog-odoo]
```
pedrobaeza, line 1: The latest copier template doesn't remove this file. On contrary, it includes more lines.

pedrobaeza (commented):
I also miss the file `.gitattributes` or similar.

checklog-odoo.cfg
eugenios73: @pedrobaeza Apologies, I have re-added the checklog-odoo.cfg file; I'm not sure why it had been deleted. The .gitattributes file is not in the commit because there haven't been any changes. I'm not entirely sure if this is correct. Is that possible?

pedrobaeza: And what about the new lines added here?

eugenios73: @pedrobaeza to ensure the process was right, I reset my local branch to the state before the changes and checked for any updates in the source branch. I re-ran update copier --trust, double-checking every prompt, but the result was identical: it deletes the checklog-odoo.cfg file and makes no changes to .gitattributes. Could you check if this is expected or if there's an issue with the copier update? I've attached screenshots of the modified files after running the command and the copier prompts.

image

image

pedrobaeza: Copier sometimes fail as it seems. The complete file should be: https://github.com/OCA/oca-addons-repo-template/blob/master/src/%7B%25%20if%20enable_checklog_odoo%20%25%7Dchecklog-odoo.cfg%7B%25%20endif%20%25%7D

Anyway, the CI right now is red.

pedrobaeza:
OK, seeing that you fix it in other PR. I'm merging this manually.

PR #725, opened 2026-01-29
[16.0][FIX] crm_date_deadline_required: fix usability issue on Kanban quick create

### LINKED ISSUE #695: [16.0] crm_date_deadline_required: Quick create form closes automatically when setting expected closing date in Kanban view
Hi @EmilioPascual ,

We’ve detected a usability issue in the crm_date_deadline_required module on version 16.0.

When using the Kanban view with quick creation of opportunities, if the first field set is the Expected Closing Date, the quick create form closes automatically before the opportunity can be completed or saved.

This behavior seems related to the required field logic added by the module. It disrupts the sales workflow and prevents users from creating opportunities efficiently.

**The detailed issue description can be found below:**

- Impacted versions:

16.0

- Steps to reproduce:

Go to the CRM → Pipeline (Kanban view).

Click "Add" to create a new opportunity using the quick create form.

Before setting any other field, set the Expected Closing Date (date_deadline).

- Current behavior:
As soon as the expected closing date is set, the quick create form closes automatically and the opportunity is not properly created.

- Expected behavior:
The quick create form should remain open after setting the expected closing date so the user can complete the other fields before saving.

- Business case / Justification:
This behavior interrupts the user flow when creating opportunities directly from the Kanban view, especially for sales teams who use quick creation frequently.

Thanks a lot for your time and great work maintaining this repo! 🙏

Best regards,

@moduon MT-12143

robperezit:
In Odoo 16.0, adding a date field to the Kanban Quick Create form causes the form to close prematurely when the datepicker loses focus due to the OWL component lifecycle. This commit moves the requirement logic to the backend to preserve usability while ensuring data integrity.

### Bug Description
As reported in #695, when using the `crm_date_deadline_required` module in Odoo 16.0, the Kanban "Quick Create" form closes automatically if the user interacts with the `date_deadline` field (DatePicker) before completing other fields. This is caused by the native OWL lifecycle in the Kanban view, which triggers a save/close event upon focus loss of date widgets.

### Solution
Since the framework's behavior in v16 makes it unstable to include a required date field directly in the Quick Create popover, this PR implements a more robust "Backend-First" approach:

1.  **Backend Validation**: Added `@api.constrains` and overrode the `create` method in `crm.lead` to ensure `date_deadline` is strictly mandatory for opportunities at the server level.
2.  **View Optimization**: Removed the `quick_create` view inheritance to prevent the UI crash/auto-close. 
3.  **Standard Workflow**: By enforcing the requirement in Python, if a user attempts to "Quick Create" an opportunity without a date, Odoo will trigger a `ValidationError`. The native web client then automatically redirects the user to the full form view with all previously entered data, where they can set the date without UI issues.

### Integrity
This change ensures that no opportunity can be created without a deadline through any channel (UI, API, or Import), while restoring a fluid workflow for sales teams using the Pipeline Kanban view.

Fixes #695

### DIFF
--- a/crm_date_deadline_required/__init__.py
+++ b/crm_date_deadline_required/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/crm_date_deadline_required/models/__init__.py
+++ b/crm_date_deadline_required/models/__init__.py
@@ -1,0 +1,1 @@
+from . import crm_lead
--- a/crm_date_deadline_required/models/crm_lead.py
+++ b/crm_date_deadline_required/models/crm_lead.py
@@ -1,0 +1,23 @@
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class CrmLead(models.Model):
+    _inherit = "crm.lead"
+
+    @api.constrains("date_deadline", "type")
+    def _check_date_deadline_required(self):
+        for record in self:
+            if record.type == "opportunity" and not record.date_deadline:
+                raise ValidationError(
+                    _("The expected closing date is required for opportunities.")
+                )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if vals.get("type") == "opportunity" and not vals.get("date_deadline"):
+                raise ValidationError(
+                    _("The expected closing date is required for opportunities.")
+                )
+        return super().create(vals_list)
--- a/crm_date_deadline_required/tests/test_crm_date_deadline_required.py
+++ b/crm_date_deadline_required/tests/test_crm_date_deadline_required.py
@@ -1,6 +1,7 @@
 # Copyright 2025 Moduon Team S.L.
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
 
+from odoo.exceptions import ValidationError
 from odoo.tests import Form
 from odoo.tests.common import TransactionCase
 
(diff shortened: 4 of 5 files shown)


robperezit:
All functional tests and linting checks are passing. Regarding the Codecov alert, the patch has 90% coverage, which ensures the new backend validation is correctly tested. The logic is ready for review.

crm_date_deadline_required/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "CRM Date Deadline Required",
     "summary": "Field date deadline required in the opportunity",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.0.1",
```
EmilioPascual: It's not necessary update version, when the PR is merged, the versión is updated.

EmilioPascual (changes requested):
@robperezit can you do squash for all commits, just 1 commit please

robperezit:
Hi @EmilioPascual, thank you for the review. I have squashed the commits into a single one and reverted the version change in the manifest as requested. All functional tests and linting checks are green. Ready for final review!

EmilioPascual (approved):
LGTM. Code review

luisDIXMIT (approved):
Code review and LGTM!

EmilioPascual:
@robperezit can you rebase and fix pre-commit? It was ko when I merged the PR.

robperezit:
Hi @EmilioPascual, the pre-commit job fails with "ModuleNotFoundError: No module named 'pkg_resources'" in the setuptools-odoo-make-default and setuptools-odoo-get-requirements hooks. This is an environment configuration issue - the runner needs setuptools installed. All functional checks (pylint, flake8) passed successfully. With 2 LGTMs, this is ready once the CI environment is fixed.

PR #727, opened 2026-02-09
[19.0][MIG] crm_industry: Migration to 19.0

muaazsiddiq:
- [x] https://github.com/OCA/partner-contact/pull/2283


crm_industry/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2015 Antiun Ingenieria S.L. - Javier Iniesta
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "CRM Industry",
+    "summary": "Link leads/opportunities to industries",
+    "version": "19.0.1.0.2",
```
luisDIXMIT: Update version to 19.0.1.0.0

luisDIXMIT (commented):
If you want to pass the tests until the dependency is merged, you can add dependency modules like this: https://github.com/OCA/partner-contact/pull/2270/changes/0588a05f5771420086daa75222019da16e70cb5a

StefanRijnhart:
Now that the dependency is merged, can you rebase please? This should fix the PR checks.

muaazsiddiq:
@StefanRijnhart 
Done

crm_industry/models/crm_lead.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+        return values
+
+    @api.onchange("partner_id")
+    def _onchange_partner_id(self):
+        if self.partner_id:
+            if self.partner_id.industry_id:
+                self.industry_id = self.partner_id.industry_id
+            if self.partner_id.secondary_industry_ids:
+                self.secondary_industry_ids = self.partner_id.secondary_industry_ids
```
StefanRijnhart: This usage an onchange plus an override of `create` to sync fields from the partner to the lead is long due for an overhaul. In Odoo, onchanges were replaced by compute methods in 14.0. Would you mind replacing this onchange by a separate onchange method for each field?

muaazsiddiq:
```Python
@api.onchange("partner_id")
    def _onchange_partner_id_industry_id(self):
        self.industry_id = False
        if self.partner_id and self.partner_id.industry_id:
            self.industry_id = self.partner_id.industry_id

    @api.onchange("partner_id")
    def _onchange_partner_id_secondary_industry_ids(self):
        self.secondary_industry_ids = [Command.clear()]
        if self.partner_id and self.partner_id.secondary_industry_ids:
            self.secondary_industry_ids = self.partner_id.secondary_industry_ids
```

@StefanRijnhart 
As per your request, this is my proposed solution.

StefanRijnhart: Apologies, I made a typo in my original comment. I wanted to say 'replacing this onchange by a separate *compute* method for each field`. Can you have another look?

muaazsiddiq: After the compute method, we don't need to override the create() method here as well. Is this correct ?

crm_industry/models/crm_lead.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class CrmLead(models.Model):
+    _inherit = "crm.lead"
+
+    industry_id = fields.Many2one(
+        comodel_name="res.partner.industry",
+        string="Main Industry",
+        compute="_compute_industry_id",
+        store=True,
```
StefanRijnhart, line 14: You'll want to make the fields `readonly=False` so that the fields can still be set by the user.

StefanRijnhart (commented):
What is the reason that the constraint method was removed?

crm_industry/models/crm_lead.py
muaazsiddiq: - After making both the fields computed, they becomes readonly=True and the computed data is coming from `partner_industry_secondary` module where the constraint is already in-check.

- If I make both the fields **readonly=False**, don't you think that the data will become inconsistent between `partner_industry_secondary` and `crm_industry` as the user will be having an option to modify the industry.

luisDIXMIT (commented):
Commits can be squashed into a single migration commit.

muaazsiddiq:
> Commits can be squashed into a single migration commit.

@luisDIXMIT I will squash the commit once the changes get finalized.

crm_industry/models/crm_lead.py
StefanRijnhart: Well, typically, a crm lead is either created from scratch and then a partner is created with the industry data gathered from the lead, or the crm lead is created for an existing partner in which case the industry data from the partner serves as a default from the lead. It is true that if the data is then modified on the lead, this will not automatically update the partner. This is the standard behaviour of several fields on the lead in the standard CRM module (e.g. the Title field), and it is consistent with the previous behaviour of the onchange mechanism in this very module.

muaazsiddiq: Ok, I will re-implement the constraint method and make the fields readonly=False

muaazsiddiq: Let me know if everything look fine now, so then i will squash the commit.

StefanRijnhart: It does look good! Please squash.

muaazsiddiq: Done

StefanRijnhart (approved):
Thanks a lot! Code review + functional test.

luisDIXMIT (approved):
Tested on runboat and LGTM!

PR #731, opened 2026-03-03
[18.0][IMP] crm_lead_to_task: Enhance task creation from leads to support optional message and attachment transfer

MohamedOsman7:
This change makes the transfer of chatter messages and attachments conditional. They are now moved only when the lead is converted (archived), while simple task creation keeps the history on the lead. This aligns the behavior with the company setting.


crm_lead_to_task/models/crm_lead.py
```diff
@@ -46,24 +46,25 @@ def _get_values_task_from_lead(self, project):
...
+                    ("res_model", "=", "crm.lead"),
+                    ("res_id", "=", self.id),
+                ]
+            ).write({"res_model": "project.task", "res_id": task.id})
 
         return task
 
     def _action_create_and_open_task(self, project):
-        task = self._create_task_from_lead(project)
+        thread = bool(self.company_id.crm_archive_lead_on_convert)
```
CRogos: This is already a boolean value. No need to convert to bool.

crm_lead_to_task/models/crm_lead.py
```diff
@@ -46,24 +46,25 @@ def _get_values_task_from_lead(self, project):
             "lead_id": self.id,
         }
 
-    def _create_task_from_lead(self, project):
+    def _create_task_from_lead(self, project, thread=False):
```
CRogos: Why did you change the method signature, and did not use the `self.company_id.crm_archive_lead_on_convert` variable?

crm_lead_to_task/__manifest__.py
```diff
@@ -11,7 +11,7 @@
     "author": "Odoo S.A., Odoo Community Association (OCA), Tecnativa",
     "website": "https://github.com/OCA/crm",
     "depends": ["crm", "project"],
-    "version": "18.0.1.1.0",
+    "version": "18.0.1.1.1",
```
CRogos: Do not change the version number. This is done automatically on merge.

CRogos (commented):
Could you have a look on my comments?

luisDIXMIT (approved):
Code review and tested on runboat, LGTM!

CRogos (approved):
LGTM.

One observation. When the lead is archived, there is no ribbon indicating this. 

Image


When we use the archive method, there is a lost ribbon which might also be not the best.

Image

https://github.com/odoo/odoo/blob/e9a8b81d4a972a57a8f46377864874190b04af2c/addons/crm/views/crm_lead_views.xml#L48-L49


Should we add an "archived" ribbon for this case?

`<widget name="web_ribbon" title="Archived" bg_color="text-bg-danger" invisible="probability == 0 and active"/>`

MohamedOsman7:
@EmilioPascual Could you merge?

EmilioPascual:
> @EmilioPascual Could you merge?

Sorry, I'm not a maintainer of this module

MohamedOsman7:
@pedrobaeza Could you merge?

pedrobaeza (commented):
Not sure you should link the archiving of the lead with the transfer of the messages/attachments.

CRogos:
I think we should keep this link. Currently there are two behaviors.
1. Convert a lead to a task, therefor also the messages/attachments are moved to the new task.
2. Create a task from this lead. This came from [crm_project_task ](https://github.com/OCA/crm/pull/667) and was merged with this module because of the big similarity. But in this case, the message/attachments should not be moved. 

So basically this is only a bugfix to restore both functionalities as before. Not sure if we need a dedicated option for not moving messages/attachments?

MohamedOsman7:
> I think we should keep this link. Currently there are two behaviors.
> 
> 1. Convert a lead to a task, therefor also the messages/attachments are moved to the new task.
> 2. Create a task from this lead. This came from [crm_project_task ](https://github.com/OCA/crm/pull/667) and was merged with this module because of the big similarity. But in this case, the message/attachments should not be moved.
> 
> So basically this is only a bugfix to restore both functionalities as before. Not sure if we need a dedicated option for not moving messages/attachments?

@pedrobaeza

pedrobaeza (approved):
/ocabot merge patch

PR #739, opened 2026-04-16
[MIG] [19.0] crm_lead_product: Migration to 19.0

PieterPaulussen:


crm_lead_product/models/crm_lead_line.py
```diff
@@ -75,10 +75,12 @@ def _onchange_product_id(self):
...
             if (
                 not self.uom_id
-                or product.uom_id.category_id.id != self.uom_id.category_id.id
+                or product.uom_id.relative_uom_id.id != self.uom_id.relative_uom_id.id
             ):
                 self.uom_id = product.uom_id.id
-            domain["uom_id"] = [("category_id", "=", product.uom_id.category_id.id)]
+            domain["uom_id"] = [
+                ("relative_uom_id", "=", product.uom_id.relative_uom_id.id)
+            ]
```
StefanRijnhart, line 83: The uom domain in Odoo 19 is based on the product's `uom_id` plus the product's `uom_ids`, as in https://github.com/odoo/odoo/blob/99c92127a8e86521f04f39d4fa775098af2af0e8/addons/stock/models/stock_move.py#L202-L205 (`seller_ids.product_uom_id` is probably less relevant in this case). Would that work well here too?

crm_lead_product/report/crm_product_report_views.xml
```diff
@@ -189,10 +189,7 @@
         >{'search_default_opportunity': True, 'search_default_current': True}</field>
         <field
             name="view_ids"
-            eval="[(5, 0, 0),
-                      (0, 0, {'view_mode': 'pivot', 'view_id': ref('crm_opportunity_report_view_pivot')}),
-                      (0, 0, {'view_mode': 'graph', 'view_id': ref('crm_opportunity_report_view_graph')}),
-                      (0, 0, {'view_mode': 'list', 'view_id': ref('crm_opportunity_report_view_tree')})]"
+            eval="[(5, 0, 0),                       (0, 0, {'view_mode': 'pivot', 'view_id': ref('crm_opportunity_report_view_pivot')}),                       (0, 0, {'view_mode': 'graph', 'view_id': ref('crm_opportunity_report_view_graph')}),                       (0, 0, {'view_mode': 'list', 'view_id': ref('crm_opportunity_report_view_tree')})]"
```
StefanRijnhart, line 192: Can you replace with `odoo.fields.Command.{clear,create}`? and the previous formatting was more readable, if pre-commit allows it.

crm_lead_product/models/crm_lead_line.py
```diff
@@ -0,0 +1,136 @@
...
+            if self.uom_id and self.uom_id.id != product.uom_id.id:
+                self.price_unit = product.uom_id._compute_price(
+                    self.price_unit, self.uom_id
+                )
+        return {"domain": domain}
+
+    @api.onchange("category_id")
+    def _onchange_category_id(self):
+        domain = {}
```
StefanRijnhart, line 92: The domain value remains empty in this method. Can it be removed?

crm_lead_product/models/crm_lead_line.py
```diff
@@ -0,0 +1,136 @@
...
+            if self.product_id and self.product_id.categ_id != categ_id:
+                self.product_id = None
+                self.name = categ_id.name
+            if self.product_tmpl_id and self.product_tmpl_id.categ_id != categ_id:
+                self.product_tmpl_id = None
+        return {"domain": domain}
+
+    @api.onchange("product_tmpl_id")
+    def _onchange_product_tmpl_id(self):
+        domain = {}
```
StefanRijnhart, line 110: Can this one be removed as well?

StefanRijnhart (commented):
Thanks! Some minor comments.

luisDIXMIT (commented):
Why is not calculating the prorated revenue?

https://github.com/user-attachments/assets/7e45153d-18e2-4ba9-a042-6149a8aa9232

crm_lead_product/models/crm_lead_line.py
PieterPaulussen: Yes, that would be better indeed. Changed to you suggestion and leaving out seller_ids.product_uom_id.

crm_lead_product/models/crm_lead_line.py
PieterPaulussen: Yes

crm_lead_product/models/crm_lead_line.py
PieterPaulussen: Yes

crm_lead_product/models/crm_lead_line.py
```diff
@@ -119,16 +117,13 @@ def _onchange_product_tmpl_id(self):
                 if self.product_id.product_tmpl_id != product_tmpl:
                     self.product_id = None
                     self.name = product_tmpl.name
-        return {"domain": domain}
 
     @api.onchange("uom_id")
     def _onchange_uom_id(self):
-        result = {}
```
PieterPaulussen, line 126: @StefanRijnhart This was also redundant similarly to the suggestions you made in the other on changes.

crm_lead_product/models/crm_lead_line.py
```diff
@@ -75,10 +75,11 @@ def _onchange_product_id(self):
                 self.name = product.name
             if (
                 not self.uom_id
-                or product.uom_id.category_id.id != self.uom_id.category_id.id
+                or product.uom_id.relative_uom_id.id != self.uom_id.relative_uom_id.id
```
StefanRijnhart, line 78: Here I would also check against allowed uoms rather than relative_uom_id. The field `relative_uom_id` is not used anywhere in business logic in Odoo 19.

StefanRijnhart (commented):
Thanks for the updates! One small change remaining, for consistency.

About the prorated revenue, looks like that is because it is not being computed for 'lead' type leads, only opportunities. Maybe that's because in some older versions, the probability field was not visible on 'lead' type leads? Maybe good to update this logic as the proability field is visible and is populated predictively nowadays.

luisDIXMIT (approved):
Code review and tested on runboat, LGTM!

PR #746, opened 2026-05-21
[19.0][FIX] crm_lead_to_task, crm_project_create: fix tests

WesleyOliveira98:
Fix error in tests

```
  File "/__w/crm/crm/crm_lead_to_task/tests/test_crm_project.py", line 33, in test_crm_project
    self.assertEqual(task.description, "<p>Description</p>")
AssertionError: Markup('<span>Description</span>') != '<p>Description</p>'
```

```
  File "/opt/odoo/odoo/tests/common.py", line 2656, in with_users
    func(self, *args, **kwargs)
  File "/__w/crm/crm/crm_project_create/tests/test_crm_create_project.py", line 63, in test_crm_create_project
    self.assertEqual(
AssertionError: Markup('<span>Test Description</span>') != Markup('<p>Test Description</p>')
```


WesleyOliveira98:
cc: @marcelsavegnago @CristianoMafraJunior @kaynnan

kaynnan (approved):

crm_project_create/tests/test_crm_create_project.py
```diff
@@ -62,5 +62,5 @@ def test_crm_create_project(self):
         self.assertEqual(self.lead.project_id.name, "Test Project")
         self.assertEqual(
             self.lead.project_id.description,
-            Markup("<p>Test Description</p>"),
+            Markup("<span>Test Description</span>"),
         )
```
BhaveshHeliconia, line 66: self.assertIn(
    "Test Description",
    self.lead.project_id.description,
)

BhaveshHeliconia (changes requested):
Instead of asserting the exact Markup value, we can make the test more robust by checking the actual content.

This is similar to the approach used in https://github.com/OCA/account-financial-tools/pull/2305, where the test was updated to verify the relevant content rather than the exact HTML/Markup representation.

PR #753, opened 2026-06-29
[16.0][ADD] crm_reminder_email_activities: send email reminder about activities that are due

alessandropecchini99:

PR #756, opened 2026-07-21
[18.0][MIG] crm_reminder_email_activities: Migration to 18.0

anusriNPS:
Migrating crm_reminder_email_activities to 18.0


anusriNPS:
image

HekkiMelody (approved):
Code and functional review, LGTM

Module was originally proposed for v16 here: https://github.com/OCA/crm/pull/753

BhaveshHeliconia (approved):

crm_reminder_email_activities/data/ir_cron.xml
```diff
@@ -0,0 +1,17 @@
...
+        <field name="name">Reminder CRM planned activities</field>
+        <field name="model_id" ref="model_res_users" />
+        <field name="active" eval="True" />
+        <field name="state">code</field>
+        <field name="code">model.cron_send_crm_reminder_activities()</field>
+        <field name="interval_number">1</field>
+        <field name="interval_type">weeks</field>
+        <field
+            name="nextcall"
+            eval="(DateTime.today() + relativedelta(weekday=6)).strftime('%Y-%m-%d')"
```
SirPyTech, line 13: **suggestion**: `strftime` might not be needed.

Moreover, why postponing the first call?

crm_reminder_email_activities/models/res_users.py
```diff
@@ -0,0 +1,27 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models
+
+
+class Users(models.Model):
+    _inherit = "res.users"
+
+    @api.model
+    def cron_send_crm_reminder_activities(self):
```
SirPyTech, line 10: **suggestion**: Please consider moving these methods to `crm.lead`: if everyone did this, the `res.users` model would be full of methods for every kind of model.

crm_reminder_email_activities/models/res_users.py
```diff
@@ -0,0 +1,27 @@
...
+        users = self._get_crm_activities().mapped("user_id")
+        mail_template = self.env.ref(
```
SirPyTech, line 12: **suggestion**: Please handle the case where the template does not exist, this might be a good opportunity to add a test for this piece of code.

crm_reminder_email_activities/models/res_users.py
```diff
@@ -0,0 +1,27 @@
...
+            "crm_reminder_email_activities.email_template_crm_reminder_activities"
+        )
+        for user in users:
+            mail_template.send_mail(user.id)
+
+    def _get_crm_activities(self, user=None):
```
SirPyTech, line 18: **suggestion**: Please consider using `self` for the user, instead of `user` since we are in the `res.users` model.

crm_reminder_email_activities/tests/test_email_reminder_planned_activity.py
```diff
@@ -0,0 +1,121 @@
+from dateutil.relativedelta import relativedelta
+
+from odoo import fields
+from odoo.tests.common import TransactionCase, new_test_user
+
+
+class TestEmailReminderPlannedActivity(TransactionCase):
```
SirPyTech, line 7: **suggestion**: We could use `BaseCommon` to reduce the mail logs as suggested in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0:
> evaluate the use of `BaseCommon` as base test class if you need to setup company, currency, users and groups in a special way)

crm_reminder_email_activities/__manifest__.py
SirPyTech, line 1: **note**: Please clarify in the PR description that this is porting an unmerged module: from the point of view of the reviewer this changes the approach.
When reviewing a migration of a merged module the reviewer just has to check that the module is correctly adapted to the target version; if the module was never merged, it has to be reviewed in full like a brand new module.

crm_reminder_email_activities/data/mail_template_reminder.xml
```diff
@@ -0,0 +1,87 @@
...
+                    </p>
+                    <ul>
+                        <t t-foreach="activities" t-as="activity">
+                            <li>
+                                <t
+                                    t-set="activity_obj"
+                                    t-value="object.env[activity.res_model].browse(activity.res_id)"
+                                />
+                                <p style="margin-bottom: 5px">Deadline:
+                                    <span t-out="activity.date_deadline" />
```
SirPyTech, line 44: **suggestion**: Consider formatting this date in the user's locale, it might suffice using `t-field` instead of `t-out`.

SirPyTech (approved):
Thanks for the PR!
This is good, I suggested some improvements but nothing worth blocking from merge in my opinion.

crm_reminder_email_activities/data/ir_cron.xml
HekkiMelody: The idea is to send the reminder on a Sunday, containing the activities for the coming week.

PR #760, opened 2026-07-29
[19.0][ADD] crm_lead_partner_by_vat: match the lead's customer by VAT

MarcGForgeFlow:
Odoo looks a lead's customer up by email. This module tries first to find a company carrying the lead's VAT (TIN), and only falls back to the native email lookup when the lead has no VAT or no company carries it.

The lookup follows the rule of account's _import_retrieve_customer_from_vat(): res.partner.vat is free text, since neither base nor base_vat normalizes it on write, so the case is never altered and both the value as typed and the one stripped of spaces and dots are searched. Exact matches keep the index on res.partner.vat and compare wildcard characters literally.

See odoo/addons/account/models/partner.py, ResPartner._import_retrieve_customer_from_vat().


MarinaAForgeFlow (approved):
LGTM

PR #364, opened 2020-11-05
14.0 mig crm phonecall

redapureskill:
**Migrating** crm_phonecall to **14.0** :+1:

PR #766, opened 2026-09-17
[18.0][FIX] crm_phonecall: context field indentation breaks module install

sergio-teruel:
The `context` text on the ir.filters records(filter_crm_phonecall_sales_team, filter_crm_phonecall_delay_to_close, filter_crm_phonecall_phone_call_to_do) was written on its own indented line. ir.filters.create()/_sanitize_shared_context() runs ast.literal_eval() on that raw text without stripping it, so the leading newline + indentation raises IndentationError and aborts module installation.

cc @Tecnativa  

ping @carlosdauden @carlos-lopez-tecnativa


carlos-lopez-tecnativa (approved):
LGTM, but I wonder how the CI passed before this change: https://github.com/OCA/crm/actions/runs/32752805081/job/97513445056#step:8:172. This was introduced during the migration to V14: https://github.com/OCA/crm/pull/364/changes/7eab872634834ebd1f2a1cb663ad1c0e08dc9b9b, and it has been present in all versions from 14 to 18, but only fails in V18. I wonder if there was any change in the core validation that caused this.

Anyway, thanks for the fix!

ValentinVinagre (commented):
LGTM :+1

ValentinVinagre (approved):
LGTM 👍🏻

carlosdauden (approved):

pedrobaeza:
The reason is this new security commit: https://github.com/odoo/odoo/commit/7e5d5fb44b5bab6c67338db7ac8795376aaa2716

But it contains a bug that was spotted by @MiquelRForgeFlow and reported, and being handled in https://github.com/odoo/odoo/pull/288528

So this shouldn't be needed tomorrow when OCB gets that bugfix.

MiquelRForgeFlow:
Can be closed.

